PR #16, opened 2020-12-01
fix: other middlewares don't run when there's an exception in fastapi-sqla context manager

vicrep:
Try to commit after the response is handled, and return a normal 500 response in that case. This allows upstream middlewares from being run (e.g. CORS), and prevents the error from being bubbled up to the top level server error handler (which bypasses other middlewares). 

I noticed this in a peculiar way btw (which is also indicative of another small logging bug IIUC):

- I have an endpoint which queries tenant from db
- endpoint raised a unhandled pydantic Validation Error somewhere
- CP (caller) gets no status code back, and a CORS error instead
- Datadog logs show “failed to commit”, even though no commits were done


vicrep:
Hello @hadrien , should we jump back onto this PR and try get it merged? It's a big QoL improvement when dealing with CORS apps, where you effectively can't get any status code on uncaught failures otherwise.

hadrien:
@vicrep CORS middleware does not run whenever an endpoint raises an unhandled exception. Therefore, I do not think this is the responsibility of fastapi-sqla to handle this. IMO it is the responsibility of the logging middleware to catch any unhandled exception and return a 500.

hadrien:
@vicrep 

```python
from fastapi import FastAPI, HTTPException
from fastapi.middleware.cors import CORSMiddleware

from dialogue.logging.fastapi import LogMiddleware

app = FastAPI()

app.add_middleware(CORSMiddleware)

@app.get("/unhandled-exception")
def raise_exception():
    raise Exception("Boom")
```

^ `/unhandled-exception` endpoint will prevent CORS middleware to run. IMO: not a `fastapi-sqla` concern but a logging middleware one.

hadrien:
> @vicrep CORS middleware does not run whenever an endpoint raises an unhandled exception. Therefore, I do not think this is the responsibility of fastapi-sqla to handle this. IMO it is the responsibility of the logging middleware to catch any unhandled exception and return a 500.

And it is the responsibility of the logging middleware to ensure middlewares run even when an unhandled exception is raised.

vicrep:
@hadrien I guess there's two facets to this.

One, that the auto-commit is really painful when it fails, as the user and app gets no feedback. I still think it should be opt-in-able, it's so far brought me more pain than benefit of saving you from writing `session.commit()`. But I digress, this is about getting feedback from the client when a request fails due to some DB (or even non-db) operation issue.

Secondly, we should be returning properly formed 500 responses in this middleware. The reason is that if you look at [Starlette's default middleware stack](https://www.starlette.io/middleware/), it has a inner-level `ExceptionMiddleware` who's job is to convert a `HttpException` to a proper response with status code. However, custom middlewares will always run _after_ this middleware (since it's the innermost one), and therefore won't benefit from the conversion of HttpException that would occur if the error was in-app. 

Can I therefore suggest then that we make an addition to this middleware so that it returns properly formed 500 responses, instead of re-raising an error? I believe that the logging middleware should already be aware / be able to understand 500 responses and log those appropriately, and I don't think that those kind of errors are "unexpected" from the PoV of this middleware, and therefore should not be raising exceptions in the middleware stack.

hadrien (approved):

PR #27, opened 2021-04-05
chore: fix tox configuration

hadrien:
## Description

* Former CI config was actually running tests against the sqlalchemy version installed by poetry;
* Changed it so that dependencies are installed by tox using `pyproject.toml`: this is the standard way IIUC;
* Today I learnt by reading on python-poetry/poetry#1941 that `[dev.dependencies]` section is not standard/compliant with [pep-517](https://www.python.org/dev/peps/pep-0517/);
* It is a `chore` as it doesn't fix anything in the library

## Related PRs

* blocking #28

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,6 @@ orbs:
   python: dialogue/python@2.12.1
   utils: dialogue/utils@3.2.6
 
-
 workflows:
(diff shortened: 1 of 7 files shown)


vicrep (approved):
TIL about that PEP. Nice one 👍

PR #28, opened 2021-04-06
feat: port to SQLAlchemy 1.4 / 2.0

hadrien:
## Description

* Pagination now allows 2.0 sqla queries.

## Related PRs

Blocked by 
* #27

### DIFF
--- a/fastapi_sqla/__init__.py
+++ b/fastapi_sqla/__init__.py
@@ -160,7 +160,7 @@ def query_count_dependency(session: Session = Depends()) -> QueryCount:
     def _query_count(query: DbQuery):  # pragma no cover
         "Dispatch on registered functions based on `query` type."
         raise NotImplementedError(
-            f"not _query_count registered for type {type(query)!r}"
+            f"no _query_count registered for type {type(query)!r}"
         )
 
     @_query_count.register
--- a/fastapi_sqla/__init__.py
+++ b/fastapi_sqla/__init__.py
@@ -5,12 +5,12 @@ from contextlib import contextmanager
 from functools import singledispatch
 from typing import Callable, Generic, List, TypeVar, Union
 
-import sqlalchemy as sa
 import structlog
 from fastapi import Depends, FastAPI, Query, Request
 from fastapi.concurrency import contextmanager_in_threadpool
 from pydantic import BaseModel, Field
 from pydantic.generics import GenericModel
+from sqlalchemy import engine_from_config
 from sqlalchemy.ext.declarative import DeferredReflection, declarative_base
 from sqlalchemy.orm import Query as LegacyQuery
 from sqlalchemy.orm.session import Session as SqlaSession
@@ -32,7 +32,7 @@ def setup(app: FastAPI):
 
 
 def startup():
-    engine = sa.engine_from_config(os.environ, prefix="sqlalchemy_")
+    engine = engine_from_config(os.environ, prefix="sqlalchemy_")
     Base.metadata.bind = engine
     Base.prepare(engine)
     _Session.configure(bind=engine)
--- a/fastapi_sqla/__init__.py
+++ b/fastapi_sqla/__init__.py
@@ -155,7 +155,7 @@ QueryCountDependency = Callable[..., QueryCount]
 PaginateSignature = Callable[[DbQuery], Page[T]]
 
 
-def query_count(session: SqlaSession, query: DbQuery) -> int:
+def query_count(session: Session, query: DbQuery) -> int:
     @singledispatch
     def _query_count(query: DbQuery):  # pragma no cover
         "Dispatch on registered functions based on `query` type."
--- a/fastapi_sqla/__init__.py
+++ b/fastapi_sqla/__init__.py
@@ -164,22 +164,15 @@ def query_count(session: Session, query: DbQuery) -> int:
     See https://gist.github.com/hest/8798884
     """
     if isinstance(query, LegacyQuery):
-        """
-        Default legacy function used to count items returned by a query.
-
-        Default Query.count is slower than a manually written query could be: It runs the
-        query in a subquery, and count the number of elements returned:
-
-        See https://gist.github.com/hest/8798884
-        """
-        return query.count()
+        result = query.count()
 
     elif isinstance(query, Select):
-        return session.execute(select(func.count()).select_from(query)).scalar()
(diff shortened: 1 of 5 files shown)


hadrien:
If you have bandwidth @vicrep 🙏🏽 ?

vicrep (commented):
Could you elaborate on why we need to make the `query_counter` a dependency? Could we not make it a _pure_ function which takes as arguments the query, which could be either a legacy or non-legacy query? 

IMO requiring you make a dependency that returns a callable as a counter function feels a little indirected from a consumer's PoV / is not a very straightforward interface.

Could we make the pagination dependency smart enough to then determine if the query is a `session.query` or a `DbQuery`, and only call `session.execute()` on the latter?

Honestly, as I think about it more, could it be worth having two separate dependencies, one to paginate classic SQLA queries, and one to paginate when passed a 1.4 `DbQuery`? That way no breaking changes, and I think there's enough overlap that most of the code would be reused between both, but that way we're super explicit.

hadrien:
> Could you elaborate on why we need to make the `query_counter` a dependency? Could we not make it a _pure_ function which takes as arguments the query, which could be either a legacy or non-legacy query?

Dependency is nice to do this:

```python
def query_count_dep(q: str = Query(...), session:Session=Depends())-> QueryCount:
    def query_count(query:DbQuery)->int:
        query = select(func.count()).where(User.name.like(q)
         return session.execute(query).scalar()
```


> IMO requiring you make a dependency that returns a callable as a counter function feels a little indirected from a consumer's PoV / is not a very straightforward interface.

I agree.

> 
> Could we make the pagination dependency smart enough to then determine if the query is a `session.query` or a `DbQuery`, and only call `session.execute()` on the latter?

yes, it is actually what that PR achieves. Out of the box, it just works with legacy and _modern_ queries. 

> 
> Honestly, as I think about it more, could it be worth having two separate dependencies, one to paginate classic SQLA queries, and one to paginate when passed a 1.4 `DbQuery`? That way no breaking changes, and I think there's enough overlap that most of the code would be reused between both, but that way we're super explicit.

I will keep the same `query_count` and remove the query count dependency thing..

That PR breaks the scoping rule. It does 2 things: porting to SQL 2.0 queries. Introducing a query count dependency. I'll lean and remove the breaking change.

vicrep:
Something similar to how the python JSON module lets you override [encoders](https://docs.python.org/3/library/json.html#json.JSONEncoder) could be nice -- when calling `json.dumps`, you can pass either: a `"default"` encoder which overrides just the `default` method of the default encoder, _or_ a `"cls"` argument to override the entire encoder class.

hadrien:
not anymore a breaking change FYI @vicrep

fastapi_sqla/__init__.py
```diff
@@ -147,24 +149,98 @@ class Page(Collection, Generic[T]):
...
+        Default Query.count is slower than a manually written query could be: It runs the
+        query in a subquery, and count the number of elements returned:
+
+        See https://gist.github.com/hest/8798884
+        """
+        return query.count()
+
+    @_query_count.register
+    def sqla14_query_count(query: Select) -> int:
+        return session.execute(select(func.count()).select_from(query)).scalar()
```
hadrien, line 179: overkill

fastapi_sqla/__init__.py
```diff
@@ -147,24 +149,98 @@ class Page(Collection, Generic[T]):
...
-    return query.count()
+def query_count(session: SqlaSession, query: DbQuery) -> int:
+    @singledispatch
+    def _query_count(query: DbQuery):  # pragma no cover
+        "Dispatch on registered functions based on `query` type."
+        raise NotImplementedError(
+            f"no _query_count registered for type {type(query)!r}"
+        )
 
+    @_query_count.register
```
vicrep, line 166:
```py
    @_query_count.register
```

looks like a display error to me (I know it's valid but that was a little confusing at first, the `@_` 😆

fastapi_sqla/__init__.py
vicrep: y overkill?

fastapi_sqla/__init__.py
```diff
@@ -147,24 +149,98 @@ class Page(Collection, Generic[T]):
...
+
+
+@paginate_query.register
+def _paginate(
+    query: Select,
+    session: Session,
+    total_items: int,
+    offset: int,
+    limit: int,
+) -> Page[T]:
```
vicrep, line 224: doesn't the generic need to be passed somehow? I feel like pydantic validation might not happen otherwise (how will it know otherwise what `T` is meant to be when doing the class instantiation? 

Check out how we do it on ER: https://github.com/dialoguemd/coredata/commit/5b5303c7cdb04ecb9d3b5c738917f1a18c98ef79

vicrep: I guess it doesn't matter for it to stay generic since fastAPI will re-parse it based on output model, but that means validation might not occur (I'm in fact very uncertain of what behaviour would be in that case)

fastapi_sqla/__init__.py
hadrien: removed

fastapi_sqla/__init__.py
hadrien: 1. IMO scope of paginate is to ... paginate, not to do any type validation; This is the role of fastapi & pydantic;
2. Passing a `response_class` to `paginate` is not acceptable IMO;
3. TIL er is not using out of the box fastapi-sqla 😢

hadrien: > 1. IMO scope of paginate is to ... paginate, not to do any type validation; This is the role of fastapi & pydantic;

Still, fastapi-sqla must allow mypy to detect static type issues

hadrien: IMO: out of scope for current PR.

hadrien: related https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html

hadrien: shall we move forward? @vicrep

vicrep: Yes, out of scope I agree. In that case we could also consider returning a dict instead of doing an extra pydantic conversion :)

vicrep (approved):
🚢 🎉

PR #30, opened 2021-04-08
chore: add optional asyncpg dependency to enable asyncio support

hadrien:
## Description

* Add optional `asyncpg` as extra

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ pytest-asyncio = { version = "^0.14.0", optional = true}
 pytest-cov = { version = "^2.10.1", optional = true}
 pytest-watch = { version = "^4.2.0", optional = true}
 tox = { version = "^3.23.0", optional = true}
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -76,6 +77,7 @@ tests = [
     "pytest-watch",
     "tox",
 ]
+asyncio = ["asyncpg"]
```
vicrep, line 80: are we assuming that sqla is a peer dependency (i.e. must be installed by consumer), or is it a true dependency? If the former, then is it fastapi-sqla's job to include this?

vicrep (approved):

pyproject.toml
hadrien: I honestly do not know. Same concern with sqlalchemy IIUC. What is your opinon?

vicrep: Let's keep it as an extra for now :) poetry / setuptools doesn't have a concept of peer dependencies AFAIK, so this is fine as long as documented

PR #31, opened 2021-04-08
feat: add support to use SQLAlchemy async sessions

hadrien:
## Description

* Add `AsyncSession` dependency;
* Add `open_async_session` async context manager;

## TODO:

~- [ ] doc in readme~ in another PR

### DIFF
--- a/fastapi_sqla/__init__.py
+++ b/fastapi_sqla/__init__.py
@@ -19,7 +19,8 @@ from sqlalchemy.sql import Select, func, select
 
 try:
     from . import asyncio_support
-    from .asyncio_support import AsyncSession, open_session as open_async_session
+    from .asyncio_support import AsyncSession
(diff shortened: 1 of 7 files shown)


vicrep (approved):
looks good!

Did bring me back to look at https://github.com/dialoguemd/fastapi-sqla/pull/16 , do you think we should get back onto that PR and add similar 'bug fix' to this PR? Otherwise you still have that pesky bug of middlewares not running on failed commits, and logs etc not being captured properly as an effect.

hadrien:
> looks good!
> 
> Did bring me back to look at #16 , do you think we should get back onto that PR and add similar 'bug fix' to this PR? Otherwise you still have that pesky bug of middlewares not running on failed commits, and logs etc not being captured properly as an effect.

yes, we must

Issue #34, opened 2021-05-02
Async dependencies of the pytest plugin are required even when you only use the sync variants


sbstp:
In my project I chose to use pyscopg2 instead of asyncpg. I did not have any issues with this until I got to the tests. I had to install `pytest-asyncio` and `asyncpg` to please the pytest fixtures.

Unfortunately even after installing the dependencies I still get an error:
```python
    @fixture(scope="session")
    def asyncpg_url(db_url):
>       scheme, parts = db_url.split(":")
E       ValueError: too many values to unpack (expected 2)
```

My connection string looks like `postgresql+psycopg2://app:*****@localhost:4200/app_test`, so there's more than 2 parts created by splitting on `':'`.

I found a work-around in the meantime, it's to override the `patch_async_sessionmaker` fixture which has `autouse=True`, with this
```python
@fixture(autouse=True)
async def patch_async_sessionmaker():
    pass
```
It doesn't cause me trouble to disable it because I only use the sync version of the fixtures.

hadrien:
@sbstp Thanks for reporting, it should be fixed by #35 
FYI I may not have a review approval before Monday end of day.

hadrien:
🎉  It should be fixed.

sbstp:
Thanks, I tested it and it indeed works well.

PR #35, opened 2021-05-02
fix: fixtures must not require asyncpg

hadrien:
## Description

As described in #34, `asynpg` is required when using pytest fixtures.

- [x] update tox configuration to allow testing with or without `asyncpg`
- [x] fix so that if `asyncpg` is not installed, it doesn't fail;
- [x] fix `async_sqlalchemy_url` to allow any URL, it currently fails if more than one colon is in the URL;


hadrien:
Regression test:

image

vicrep (approved):

PR #36, opened 2021-05-13
fix: allow paginating queries that return non scalars - DIA-29942

hadrien:
## Description

* Allow paginating queries that do not return scalars;

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ By default:
 
 #### Paginating non-scalar results
 
-When query doesn't returned scalars, specify `scalars=False` when invoking `paginate`:
+To paginate a query which doesn't return scalars, specify `scalars=False` when invoking `paginate`:
 
 ```python
 from fastapi import APIRouter, Depends
--- a/README.md
+++ b/README.md
@@ -143,6 +143,9 @@ class UserModel(BaseModel):
     id: int
     name: str
 
+    class Config:
+        orm_mode = True
+
 
 @router.get("/users", response_model=Page[UserModel])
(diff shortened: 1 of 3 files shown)


hadrien:
@vicrep As you know, I am not fond of using booleans, but, I do not have a good understanding on all types of query it would have to handle ... so good enough for now.

README.md
```diff
@@ -177,6 +180,46 @@ By default:
     }
     ```
 
+#### Paginating non-scalar results
+
+When query doesn't returned scalars, specify `scalars=False` when invoking `paginate`:
```
vicrep, line 185:
```suggestion
To paginate a query which doesn't return scalars, specify `scalars=False` when invoking `paginate`:
```

Also, is scalar a known/common enough concept, or should we perhaps link to associated docs here?

fastapi_sqla/__init__.py
```diff
@@ -243,13 +245,16 @@ def _paginate(
     total_items: int,
     offset: int,
     limit: int,
+    *,
+    scalars: bool = True,
```
vicrep, line 249: Should this be singular `scalar`? (I'm not sure, I'm asking you @hadrien )

vicrep: JK, I got confused between the old `.scalar()` from the legacy queries and `Results.scalars()`

vicrep (approved):

fastapi_sqla/__init__.py
hadrien: it is plural as it return multiple scalar.

PR #37, opened 2021-05-13
feat: use query_count as a FastAPI dependency

hadrien:
## Description

* Custom pagination relying on a callable rather than a FastAPI dependency is sub-optimal;
* This PR introduces a breaking change to change that callable to be a dependency;

### DIFF
--- a/tests/test_pagination.py
+++ b/tests/test_pagination.py
@@ -1,4 +1,3 @@
-import math
 from typing import List
 
 import httpx
@@ -9,7 +8,6 @@ from pydantic import BaseModel
 from pytest import fixture, mark, param
 from sqlalchemy import MetaData, Table, func, select, text
 from sqlalchemy.orm import joinedload, relationship
-from sqlalchemy.sql import Select
 
 
 @fixture(scope="module", autouse=True)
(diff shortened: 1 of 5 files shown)


tests/test_pagination.py
```diff
@@ -217,15 +216,18 @@ def sqla_14_all_users_with_notes_count(paginate: Paginate = Depends()):
         )
         return paginate(query, scalars=False)
 
-    def query_count(session: Session, query: Select) -> int:
-        return session.execute(select(func.count()).select_from(user_cls)).scalar()
+    def count_user_notes(user_id: int, session: Session = Depends()) -> int:
+        return session.execute(
+            select(func.count(note_cls.id)).where(note_cls.user_id == user_id)
+        ).scalar()
```
hadrien, line 222: @vicrep I strongly believe the query counter function should be a FastAPI dependency to support this ^ 

What do you think?

vicrep: @hadrien I agree in the context of SQLA2 language. The nice thing though is that making it a dependency still allows us to do pure counter functions if DI is not needed (basically, if you don't need it to be a dependency, then pretend it isn't one :) )

tests/test_pagination.py
```diff
@@ -217,15 +216,18 @@ def sqla_14_all_users_with_notes_count(paginate: Paginate = Depends()):
...
 
-    CustomPaginate: PaginateSignature = Pagination(query_count=query_count)
+    CustomPaginate: PaginateSignature = Pagination(query_count=count_user_notes)
```
vicrep, line 224: can we type the return type of `Pagination` so that `PaginateSignature` is inferred?

vicrep (approved):
Me gusta

PR #43, opened 2022-02-08
feat: support AWS IAM authentication for RDS (DIA-39107)

ecotg:
## Description 
This adds the support for AWS RDS IAM Authentication. This works with a refresh token that is used as the password and is rotated as needed i.e. depending on the session duration.

Consideration that were made here:
1. The AWS RDS IAM authentication is opt-in
2. The AWS region is not passed anywhere as `AWS_DEFAULT_REGION` is automatically picked by boto if defined
3. The database host, user, port and this is already parsed automatically by sqlalchemy based on `sqlalchemy_url` environment variable

## Related JIRA issues

* [DIA-39107]

## Related PRs

[DIA-39107]: https://dialoguemd.atlassian.net/browse/DIA-39107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fastapi_sqla/utils.py
```diff
@@ -0,0 +1,21 @@
+import os
+
+import boto3
+
+ENDPOINT = os.getenv("RDS_ENDPOINT", "")
```
mbelang, line 5:
```suggestion
ENDPOINT = os.getenv("RDS_ORIGIN", "")
```

fastapi_sqla/utils.py
```diff
@@ -0,0 +1,21 @@
...
+REGION = os.getenv("AWS_REGION", "")
+USERNAME = os.getenv("USERNAME", "")
```
mbelang, line 7:
```suggestion
USERNAME = os.getenv("SQLA_USERNAME", "")
```

Here I think we should use the same concept we did for GUESTLIST and use variables with prefix that are automatically loaded by the lib

fastapi_sqla/utils.py
```diff
@@ -0,0 +1,21 @@
+import os
+
+import boto3
+
+ENDPOINT = os.getenv("RDS_ENDPOINT", "")
+REGION = os.getenv("AWS_REGION", "")
```
mbelang, line 6:
```suggestion
REGION = os.getenv("AWS_DEFAULT_REGION", "")
```
Default region is the one injected by the cluster already and normally picked up by BOTO automatically

mbelang:
@edouardlp I think we will need to brainstorm a bit here to make that work with Istio security and all.

fastapi_sqla/utils.py
```diff
@@ -0,0 +1,36 @@
+import os
```
hadrien, line 1: Is this module actually `fastapi_sqla.rds_support`?

pyproject.toml
```diff
@@ -57,6 +57,7 @@ pytest-asyncio = { version = "^0.14.0", optional = true}
 pytest-cov = { version = "^2.10.1", optional = true}
 pytest-watch = { version = "^4.2.0", optional = true}
 tox = { version = "^3.23.0", optional = true}
+boto3 = "^1.20.50"
```
hadrien, line 60:
```suggestion
boto3 = { version = "^1.20.50", optional = true }
```

tests/test_startup.py
```diff
@@ -71,3 +71,34 @@ async def test_async_startup_fail_on_bad_async_sqlalchemy_url(monkeypatch):
...
+@mark.asyncio
+@mark.dont_patch_engines
+async def test_async_startup_with_dynamic_password(monkeypatch, async_sqlalchemy_url):
+    from fastapi_sqla.asyncio_support import startup
+
+    monkeypatch.setenv("aws_rds_iam_enabled", True)
+    monkeypatch.setenv("async_sqlalchemy_url", async_sqlalchemy_url)
+
+    with patch(
+        "fastapi_sqla.aws_rds_iam_support.get_authentication_token", return_value=None
```
mbelang, line 100: I will mock boto call instead of internal function.

fastapi_sqla/utils.py
mbelang: Yes I renamed it. Make total sense.

pyproject.toml
mbelang: Fixed.

fastapi_sqla/_pytest_plugin.py
```diff
@@ -26,7 +26,7 @@ def db_url():
     Else, host used is `localhost`: postgresql://postgres@localhost/postgres
     """
     host = "postgres" if "CI" in os.environ else "localhost"
-    return f"postgresql://postgres@{host}/postgres"
+    return f"postgresql://postgres@{host}:5432/postgres"
```
edouardlp, line 29: Rather than requiring the db_url to always have a port (which I'm assuming is why you added this), could we instead default to 5432 [here](https://github.com/dialoguemd/fastapi-sqla/pull/43/files#diff-b2ab2a53b6ed37e0cc84a6d09670f9e1fb169ba79fb25898d06ab9ce56776415R17)?

fastapi_sqla/aws_rds_iam_support.py
```diff
@@ -0,0 +1,25 @@
+import boto3
+from pydantic import BaseSettings
+
+
+def get_authentication_token(host, port, user):
+    session = boto3.Session()
+    client = session.client("rds")
```
edouardlp, line 7: What's involved in creating the client? Is it cheap? If it's not it _might_ be worth caching it.

fastapi_sqla/aws_rds_iam_support.py
```diff
@@ -0,0 +1,25 @@
...
+        cparams["password"] = get_authentication_token(
+            host=cparams["host"], port=cparams["port"], user=cparams["user"]
+        )
+
+
+class Config(BaseSettings):
+    aws_rds_iam_enabled: bool = False
+
+    def __init__(self):
+        return super().__init__()
```
edouardlp, line 25: Is this really needed?
```suggestion
```

fastapi_sqla/aws_rds_iam_support.py
```diff
@@ -0,0 +1,25 @@
...
+
+    if config.aws_rds_iam_enabled:
+        cparams["password"] = get_authentication_token(
+            host=cparams["host"], port=cparams["port"], user=cparams["user"]
+        )
+
+
+class Config(BaseSettings):
+    aws_rds_iam_enabled: bool = False
+
```
edouardlp, line 23: Do you think we should use a prefix like `fastapi_sqla`?

```
class Config:
        env_prefix = 'fastapi_sqla_'
```

fastapi_sqla/_pytest_plugin.py
mbelang: I quickly checked and we always pass the port in the sql_alchemy variable. What if we need to override? I suggest we get the port if it is set, otherwise we take 5432. What do you think?

fastapi_sqla/aws_rds_iam_support.py
mbelang: Good question. Since it might be called often, caching is probable worth it.

fastapi_sqla/aws_rds_iam_support.py
edouardlp: I think in the case of hello-k8s that was to make it work with `Depends`, but you don't need that here.

fastapi_sqla/_pytest_plugin.py
edouardlp: Yes that sounds great

mbelang: Done

fastapi_sqla/aws_rds_iam_support.py
mbelang: Done

fastapi_sqla/__init__.py
```diff
@@ -63,6 +65,8 @@ def setup(app: FastAPI):
 def startup():
     lowercase_environ = {k.lower(): v for k, v in os.environ.items()}
     engine = engine_from_config(lowercase_environ, prefix="sqlalchemy_")
+    event.listen(engine.engine, "do_connect", set_connection_token)
```
hadrien, line 68: This should be in the `fastapi_sqla.aws_rds_iam_support` module as it is irrelevant if rds support is not needed.

fastapi_sqla/aws_rds_iam_support.py
hadrien:
```
❯ python -m timeit -n 1000 'import boto3;session = boto3.Session();client = session.client("rds")'
1000 loops, best of 5: 55.8 msec per loop
❯ python -m timeit -n 1000 'dict(a=1, b=2, c=3, d=4)'                                                                                                                                                  
1000 loops, best of 5: 161 nsec per loop
```

Takes at best 55ms.

~[ ](https://media.giphy.com/media/l0MYNI50nCCOa3f8c/giphy.gif)

hadrien: But IIUC, it is only called when connecting.

hadrien: Which occurs at startup.

edouardlp: I believe it is called whenever a new connection to the DB is established which would occur during the lifetime of the service as well as more DB connections are needed or as existing ones are disconnected for any number of reasons.

edouardlp: 55ms to establish a new connection is not great, but not a big deal. Since caching correctly is hard it's probably best to leave it as is for now and readjust if we se a problem.

hadrien: > I believe it is called whenever a new connection to the DB is established

True. 
By default sqla connection pool size is 5 with a overflow max of 10. Also, by default, all connections are recycled.

https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine

pyproject.toml
```diff
@@ -57,6 +57,7 @@ pytest-asyncio = { version = "^0.14.0", optional = true}
 pytest-cov = { version = "^2.10.1", optional = true}
 pytest-watch = { version = "^4.2.0", optional = true}
 tox = { version = "^3.23.0", optional = true}
+boto3 = { version = "^1.20.50", optional = true }
```
hadrien, line 60: Add it to extras as well?

fastapi_sqla/aws_rds_iam_support.py
mbelang: We should probably bump the default a bit as it bites us in the past but problem for later I suppose.

mbelang:
> Code design proposal:
> 
>     1. Write `fastapi_sqla.aws_rds_iam_support:startup` function which setup whatever needed in both sync and async mode (if enabled)
> 
>     2. Update `fastapi_sqla.setup` to register `fastapi_sqla.aws_rds_iam_support:startup` function if enabled
> 
>     3. Remove any import to `fastapi_sqla.aws_rds_iam_support` anywhere else

@hadrien I'm looking at this and I see that I will need to duplicate some logic regarding the sync/async stuff. All that the rds iam support needs is the engine.

Would you be in agreement with this:
1) Create a fastapi_sqla.aws_rds_iam_support:startup(engine) function
2) Make the check in the function to see if it is enable or not
3) Make the event call in the startup function

I'm not certain I see the overall improvement in the proposal you make. 
Do you have any other suggestion that would make the code simple and DRY?

pyproject.toml
```diff
@@ -76,6 +77,7 @@ tests = [
     "pytest-cov",
     "pytest-watch",
     "tox",
+    "boto3",
 ]
 asyncpg = ["asyncpg"]
```
edouardlp, line 82: You'll need to declare another extra so that client projects can declare the extra and get the right version of boto installed, something like
`aws_rds_iam` = ["boto3"]`

fastapi_sqla/aws_rds_iam_support.py
```diff
@@ -0,0 +1,30 @@
+import boto3
+from pydantic import BaseSettings
+from sqlalchemy import event
+
+
+def startup(engine):
```
edouardlp, line 6: I do care 1/10 but it's not really starting anything
```suggestion
def configure(engine):
```

pyproject.toml
mbelang: Yes I was doing that ATM :)

fastapi_sqla/aws_rds_iam_support.py
mbelang: I called it setup to be consistent with the lib.

pyproject.toml
```diff
@@ -78,7 +79,7 @@ tests = [
     "tox",
```
edouardlp, line 79: Why was boto3 removed from the `tests` extra? Isn't it required to run the tests?

mbelang: This is right but asyncpg is also required for the tests and it is not in the tests section

mbelang: But I see that the test will be skipped if asyncpg is not installed

fastapi_sqla/_pytest_plugin.py
```diff
@@ -17,16 +17,33 @@
...
 @fixture(scope="session")
-def db_url():
-    """Default db url used by depending fixtures.
+def db_host():
+    """Default db host used by depending fixtures.
+
+    When CI key is set in environment variables, it uses `postgres` as host name else, host used is `localhost`
+    """
+
+    return "postgres" if "CI" in os.environ else "localhost"
```
edouardlp, line 26: Why not check if a `PYTEST_DB_HOST` variable is set and set `postgres` in the CI? Some CI configurations may not be able to offer a `postgres` host so this seems more flexible and less convoluted.

fastapi_sqla/aws_rds_iam_support.py
```diff
@@ -0,0 +1,30 @@
+import boto3
```
edouardlp, line 1: Since `boto3` is optional, this will fail to import on systems that don't have it. We have to import it in a try/except.

I suggest that when boto3 is not set we also raise on [this line](https://github.com/dialoguemd/fastapi-sqla/pull/43/files#diff-b2ab2a53b6ed37e0cc84a6d09670f9e1fb169ba79fb25898d06ab9ce56776415R10) with a message explaining that boto3 must be installed.

edouardlp: [This approach looks good to me](https://github.com/dialoguemd/fastapi-sqla/blob/master/fastapi_sqla/__init__.py#L30)

fastapi_sqla/_pytest_plugin.py
mbelang: This logic was already there. I only moved it and fixing this is out of scope of this PR.

Personally even if we check of PTTEST_DB_HOST we still have to check of local vs ci because locally the host is called `localhost` and on CI it is called `postgres` because docker.

fastapi_sqla/aws_rds_iam_support.py
mbelang: Done

fastapi_sqla/_pytest_plugin.py
edouardlp: > out of scope of this PR

Fair

> still have to check of local vs ci 

What I meant is defaulting to localhost but using `PYTEST_DB_HOST` if it exists

mbelang: Right. Good idea

tests/conftest.py
```diff
@@ -95,6 +107,14 @@ def check_asyncpg(request):
         skip("This test requires asyncpg. Skipping as asyncpg is not installed.")
 
 
+@fixture(autouse=True)
+def check_bobo3(request):
+    "Skip test marked with mark.require_boto3 if boto3  is not installed."
+    marker = request.node.get_closest_marker("require_boto3")
+    if marker and not is_boto3_installed():
+        skip("This test requires boto3. Skipping as asyncpg is not installed.")
```
edouardlp, line 115: 🍝 
```suggestion
        skip("This test requires boto3. Skipping as boto3 is not installed.")
```

tests/test_startup.py
```diff
@@ -71,3 +84,41 @@ async def test_async_startup_fail_on_bad_async_sqlalchemy_url(monkeypatch):
...
+@mark.require_boto3
+@mark.dont_patch_engines
+def test_sync_startup_with_aws_rds_iam_enabled(
+    monkeypatch, boto_session, boto_client, db_host, db_user
+):
+    from fastapi_sqla import startup
+
+    monkeypatch.setenv("fastapi_sqla_aws_rds_iam_enabled", True)
+
+    with patch("boto3.Session", return_value=boto_session):
```
edouardlp, line 98: Nit but why not patch in the `boto3_session` fixture?

tests/conftest.py
mbelang: done

tests/test_startup.py
```diff
@@ -16,6 +16,19 @@ def case_sensitive_environ(environ, request):
...
+@fixture()
+def boto_session(boto_client):
+    boto_session_mock = Mock()
+    boto_session_mock.client.return_value = boto_client
+
+    return boto_session_mock
+
+
+@fixture()
+def boto_client():
```
edouardlp, line 28:
```suggestion
def boto_client_mock():
```

tests/test_startup.py
```diff
@@ -16,6 +16,19 @@ def case_sensitive_environ(environ, request):
         yield values
 
 
+@fixture()
+def boto_session(boto_client):
```
edouardlp, line 20:
```suggestion
def boto_session_mock(boto_client_mock):
```

hadrien:
> @hadrien I'm looking at this and I see that I will need to duplicate some logic regarding the sync/async stuff. All that the rds iam support needs is the engine.
> 
> Would you be in agreement with this:
> 
> 1. Create a fastapi_sqla.aws_rds_iam_support:startup(engine) function
> 2. Make the check in the function to see if it is enable or not
> 3. Make the event call in the startup function
> 
> I'm not certain I see the overall improvement in the proposal you make. Do you have any other suggestion that would make the code simple and DRY?

Reading the flow is simpler 1 entry point being called once and only once rather than having to read both startup to understand `setup` is called twice.

I am not fond of having unrelated function & module in charge of calling the rds setup. I'd rather have just `fastapi_sqla:setup` be in charge of registering one startup event handler in rds if enabled. 

Having `fastapi_sqla.aws_rds_iam_support:startup` calling twice a function `setup_engine` is perfectly find with me. 

I do not see where it DRYs anything?

hadrien:
> I do not see where it DRYs anything?

Ok as it'd have to verify async support twice.

Go ahead, I'll propose something later on.

tests/test_startup.py
mbelang: done

tests/test_startup.py
mbelang: done

tests/test_startup.py
mbelang: Done

pyproject.toml
hadrien: It is not required in the default configuration so that tox can run with or without it.

hadrien (approved):

pyproject.toml
edouardlp: Yeah we just figured it out

mbelang:
Since I change the coverage flags, codecov is not able to do the diff with master but the diff is 100% on patch.

edouardlp (approved):

PR #46, opened 2022-02-10
fix: aws_iam_rds_support independant from startup event handlers

hadrien:
## Description 

* `aws_rds_iam_support` does not depend anymore on unrelated startup handlers;
* Move config in top level module
* Have a separated file for `aws_rds_iam_support` tests

## Related JIRA issues

* [DIA-XXXXX]

### DIFF
--- a/fastapi_sqla/aws_rds_iam_support.py
+++ b/fastapi_sqla/aws_rds_iam_support.py
@@ -10,19 +10,21 @@ from sqlalchemy import event
 from sqlalchemy.engine import Engine
 
 
-def startup():
+def setup():
     assert boto3_installed, boto3_installed_err
-    event.listen(Engine, "do_connect", set_connection_token)
+    event.listen(Engine, "do_connect", _set_connection_token)
 
 
-def get_authentication_token(host, port, user):
+def tear_down():
+    event.remove(Engine, "do_connect", _set_connection_token)
+
+
+def _set_connection_token(dialect, conn_record, cargs, cparams):
     session = boto3.Session()
     client = session.client("rds")
-    token = client.generate_db_auth_token(DBHostname=host, Port=port, DBUsername=user)
-    return token
-
-
-def set_connection_token(dialect, conn_rec, cargs, cparams):
-    cparams["password"] = get_authentication_token(
-        host=cparams["host"], port=cparams.get("port", 5432), user=cparams["user"]
+    token = client.generate_db_auth_token(
+        DBHostname=cparams["host"],
+        Port=cparams.get("port", 5432),
+        DBUsername=cparams["user"],
     )
+    cparams["password"] = token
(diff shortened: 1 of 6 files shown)


hadrien:
What I failed explaining yesterday.

hadrien:
@mbelang 

> Cool. Can you test it in notepad please and make the result of the tests in a Validation section of the PR please?

Do you mean running notepad tests with this PR?

> If the pods starts without crash looping that it is a success

I don't how to deploy notepad dev with that PR (without working on dev publishing).

(Nothing urgent BTW)

mbelang:
> Do you mean running notepad tests with this PR?

No I mean doing what I did yesterday. I publish fastapi-sqla on gemfury with the pattern we use for dev published libs `0.dev+commit-long-sha` (I did it from my machine)

Then in notepad, I used that version in pyproject.toml and push a commit.
Since notepad has ephemeral deploy, you will be able to test this in isolation

You can do all of that OR you can add the dev publish on our private gemfury repo (dunno if pypi supports dev publish) and then no more manual steps

hadrien:
@mbelang dialoguemd/notepad#23 seems to deploy without any issue.

fastapi_sqla/aws_rds_iam_support.py
```diff
@@ -6,33 +6,25 @@
...
     boto3_installed_err = str(err)
 
-from pydantic import BaseSettings
 from sqlalchemy import event
+from sqlalchemy.engine import Engine
 
 
-def setup(engine):
-    config = Config()
+def setup():
```
mbelang, line 13: will [this](https://github.com/dialoguemd/notepad/pull/22/files#diff-4a04c50b1e50e5ad153b61f8fd77103944ce3d94717ccc30f19ed4086caee9a2R40) still work without the engine?

hadrien: I believe it will. let me check.

hadrien: Yes, `connectable` is an instance of `sqlalchemy.engine.Engine`, and this PR makes the code tio listen to all `"do_connect"` event from any `Engine` instance

edouardlp: Won't this cause problems if someone is using multiple sqlalchemy engines, such as having an RDS engine, and maybe a sqlite or something for caching?

Is it possible that we may have prioritized aesthetics over functionality. 

Since I don't foresee Dialogue ever running into this scenario I do not mind this PR at all.

hadrien:
> Won't this cause problems if someone is using multiple sqlalchemy engines, such as having an RDS engine, and maybe a sqlite or something for caching?

yes definitely

> Is it possible that we may have prioritized aesthetics over functionality?

I agree

> Since I don't foresee Dialogue ever running into this scenario I do not mind this PR at all.

_Le mieux est l'ennemi du bien._

PR #53, opened 2022-07-11
fix: cache rds client [DIA-44302]

mbelang:
## Description 
Cache the rds client to speed up connection creation. Only the first connection to be created will have a hit on performance and the others will be faster to create.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,7 +110,7 @@ executors:
   python-postgres:
     docker:
       - image: cimg/python:3.9
-      - image: postgres:10.12
+      - image: postgres:14.4
         name: postgres
         environment:
           POSTGRES_HOST_AUTH_METHOD: trust
--- a/fastapi_sqla/aws_rds_iam_support.py
+++ b/fastapi_sqla/aws_rds_iam_support.py
@@ -6,6 +6,8 @@ except ImportError as err:
     boto3_installed = False
     boto3_installed_err = str(err)
 
+from functools import cache
+from this import d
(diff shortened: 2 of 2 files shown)


dialogue-shreyas (approved):

vicrep (approved):
LGTM! I wonder if we should consider generating the initial client / connection as part of the startup sequence, so that the "slow" first call is not randomly given to an unlucky request. WDYT?

mbelang:
> LGTM! I wonder if we should consider initializing (and caching) the initial client / connection as part of the startup sequence, so that the "slow" first call is not randomly given to an unlucky request. WDYT?

That is a really good suggestion. Since the function caches the client, we could simply call the get_rds_client between those 2 [lines](https://github.com/dialoguemd/fastapi-sqla/blob/master/fastapi_sqla/aws_rds_iam_support.py#L17-L18)? Feels like a hack though

vicrep:
@mbelang I don't think it would be that hacky, since that method is called by the library's `startup` which is called on app startup. I would either have put it where you suggested, or directly in the `startup` method, but I think your suggestion made the most sense.

mbelang:
> @mbelang I don't think it would be that hacky, since that method is called by the library's `startup` which is called on app startup. I would either have put it where you suggested, or directly in the `startup` method, but I think your suggestion made the most sense.

Perfect then it is done now.

mbelang:
Semantic is broken again will merge with admin rights

vicrep (approved):

PR #54, opened 2022-07-29
fix: remove autoused fixtures + set dont_patch_engine marker in plugin - DM-6561

hadrien:
## Description 

* The _autoused_ fixtures force a connection to postgres server even if the test being run does not require a connection to postgres;
* The `dont_patch_engine` marker is now set by plugin rather than in conftest;
* Removed unused fixture `sqla_version_tuple`;

## Related JIRA issues

* [DM-6561]

## Related PRs

* dialoguemd/engage#22

[DM-6561]: https://dialoguemd.atlassian.net/browse/DM-6561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fastapi_sqla/_pytest_plugin.py
+++ b/fastapi_sqla/_pytest_plugin.py
@@ -16,11 +16,8 @@ except ImportError:
     asyncio_support = False
 
 
-@fixture(scope="session")
-def sqla_version_tuple():
-    from sqlalchemy import __version__
-
-    return tuple(int(i) for i in __version__.split("."))
+def pytest_configure(config):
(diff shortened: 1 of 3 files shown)


edouardlp (approved):
Thanks! Check codecov, it seems that this change is leading some fixtures to no longer being executed as part of this repo's tests and thus drops coverage.

guillaumeraymond (approved):

PR #76, opened 2023-02-01
feat: add compatibility with sqlalchemy 2.0 - DIA-51295

hadrien:
## Description 

🎉 [SQLAlchemy 2.0.0 has been released on January 26th](https://www.sqlalchemy.org/blog/2023/01/26/sqlalchemy-2.0.0-released/) 

## Related

* [DIA-51295]


hadrien:
@vicrep 
approve pr

tests/test_middleware.py
```diff
@@ -134,10 +134,13 @@ async def test_async_session_dependency(client, faker, async_session):
 
 @fixture
 def user_1(sqla_connection):
-    sqla_connection.execute(
-        text("INSERT INTO public.user VALUES (1, 'bob', 'morane') ")
-    )
+    with sqla_connection.begin():
```
vicrep, line 137: Would this be a breaking change for consumers of the library that aren't using v2 or is this whole thing transparent to users? Only asking because tests had to be modified here.

vicrep (approved):
@hadrien sorry I forgot to press "approve" 🙈 LGTM :shipit:

tests/test_middleware.py
hadrien: I am not sure I understand your concern

Unmodified tests using that fixture were still passing with sqla1.3, 1.4. 
Without that change, in sqla 2.0, test would deadlock (1 active transaction locked from another). 

Last, The tear down was not done, but it is the right thing to do.

hadrien: TL; DR: There is no breaking change.

PR #79, opened 2023-02-09
feat(aurora): invalidate connection on readonly transaction error - DIA-51197

arththebird:
## Description 

Invalidate the database connection when receiving an error caused by trying to execute a write statement on a readonly connection. This will help mitigate issues with AWS Aurora, caused by errors during failover events that can result in the write endpoint pointing to a read instance.

Note that this is an opt-in behaviour, enabled by setting the `fastapi_sqla_aws_aurora_enabled` environment variable to `true`.

## Related JIRA issues

* [DIA-51197]

## Validation

Validated in tests by using a readonly connection ✅ 

[DIA-51197]: https://dialoguemd.atlassian.net/browse/DIA-51197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/test_aws_aurora_support.py
+++ b/tests/test_aws_aurora_support.py
@@ -55,3 +55,14 @@ def test_readonly_error_codes(pgcode, should_disconnect):
     disconnect_on_readonly_error(context)
 
     assert context.is_disconnect == should_disconnect
+
+
+def test_already_disconnected():
+    from fastapi_sqla.aws_aurora_support import disconnect_on_readonly_error
+
+    context = Mock()
+    context.is_disconnect = True
+
+    disconnect_on_readonly_error(context)
+
+    assert context.is_disconnect is True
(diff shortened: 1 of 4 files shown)


vtremblay (approved):

tests/test_aws_aurora_support.py
```diff
@@ -0,0 +1,57 @@
+from unittest.mock import Mock
+
+from pytest import mark, raises
+from sqlalchemy import text
+
+
+@mark.sqlalchemy("1.4")
```
arththebird, line 7: `postgresql_readonly` is only available in SQLAlchemy >=1.4. [Ref](https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#setting-read-only-deferrable)

hadrien:
Should usage of aurora be documented?

PR #81, opened 2023-02-11
fix: table reflection with async db url - DIA-51295

hadrien:
## Description 

* Table reflection does not work when using async sqlalchemy configuration with `sqlalchemy>=2.0.0`;
* It was not discovered by tests: I noticed it when trying to upgrade a project to `fastapi-sqla=2.8.0` and   `sqlalchemy>=2.0.0`;
* It worked in tests because the fixture `async_session` depended on `sqla_reflection` which would run reflection in synchronous mode; Any subsequent call to `Base.prepare` to do reflection would do nothing as reflection already occurred in test setup;
* Added a `async_sqla_reflection` fixture to make async tests fail;
* Added an integration folder in tests and added a set of test in async mode; Will continue in other PRs to reorganize tests;
* Updated tearing down configured sqla mappers & metadata in tests;
* Renamed tables created in test to ease troubleshooting;

### TODO:

* [x] Reproduce bug in test
* [x] Fix it

## Related

* #76 
* [DIA-51295]

### DIFF
--- a/fastapi_sqla/_pytest_plugin.py
+++ b/fastapi_sqla/_pytest_plugin.py
@@ -92,10 +92,9 @@ def sqla_modules():
 
 
 @fixture
-def sqla_reflection(sqla_modules, sqla_connection, db_url):
+def sqla_reflection(sqla_modules, sqla_connection):
     import fastapi_sqla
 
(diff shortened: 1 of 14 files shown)


hadrien:
Closing to simplify in another PR

PR #82, opened 2023-02-20
fix: table reflection with async db url - DIA-51295

hadrien:
## Description 

* Table reflection does not work when using async sqlalchemy configuration with sqlalchemy>=2.0.0;
* It was not discovered by tests: I noticed it when trying to upgrade a project to fastapi-sqla=2.8.0 and sqlalchemy>=2.0.0;
* It worked in tests because the fixture async_session depended on sqla_reflection which would run reflection in synchronous mode; Any subsequent call to Base.prepare to do reflection would do nothing as reflection already occurred in test setup;
* Updated `asyncio_support` startup & added a async_sqla_reflection fixture to make async tests fail;

I want to reorganize tests and introduce real integration tests. Doing it in current PR would make it too big IMO;

## Related

* [DIA-51295]
* #76
* Closed #81 in favor of current PR.

### DIFF
--- a/fastapi_sqla/_pytest_plugin.py
+++ b/fastapi_sqla/_pytest_plugin.py
@@ -178,7 +178,15 @@ if asyncio_support:
                 yield new_engine
 
     @fixture
-    async def async_session(async_sqla_connection, sqla_reflection, patch_new_engine):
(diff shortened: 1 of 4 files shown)


vicrep (approved):
Looks good!

PR #90, opened 2023-04-05
feat: add support for overloading Base class (proposal)

vicrep:
## Description 

FastAPI-sqla currently enforces we use the provided `Base` class, such as preventing consumers from using non-reflected base classes (such as the new SQLA2 `DeclarativeBase` class).

This proposal adds the ability to pass in a custom `base_cls` to the `fastapi_sqla.setup()` function, in order to allow more flexibility.

e.g.
```py
# --------------------my_app/sqla.py------------------- #
from sqlalchemy.orm import DeclarativeBase

class Base(DeclarativeBase):
  __abstract__ = True


# --------------------my_app/main.py------------------- #
from fastapi_sqla import setup()

from my_app.sqla import Base

app = FastAPI()

setup(app, base_cls=Base)
```

If this proposal sounds acceptable, then I'll work on writing documentation on the matter inc. updating the pytest plugin to support (or document how to support) passing in a custom base.

Note: tests fail due to assertion that's a little strict on `app.add_event_handler` being called, but this should be fully backwards compatible.

### DIFF
--- a/fastapi_sqla/__init__.py
+++ b/fastapi_sqla/__init__.py
@@ -1,4 +1,6 @@
+from functools import partial
 import os
+from typing import Union
 
 from fastapi import FastAPI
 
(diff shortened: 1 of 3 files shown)


vicrep:
Closing as I don't think this makes sense yet... declarative bases don't need to be "prepared" so we can just create our own bases in-repo if needed. Might open a PR later to disable reflection though (though it has no impact but is kinda redundant)

PR #99, opened 2023-08-18
chore(ci): use new contexts - DIA-59103

arththebird:
## Description 

Remove gemfury context in favor of new contexts.

Update orbs to latest.

## Related JIRA issues

* [DIA-59103]

## Validation

Ci still works ✅ 

[DIA-59103]: https://dialoguemd.atlassian.net/browse/DIA-59103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,31 +1,35 @@
 version: 2.1
 
 orbs:
-  base: dialogue/base@1.17.1
-  release: dialogue/release@2.12.1
-  python: dialogue/python@3.11.1
-  utils: dialogue/utils@3.12.1
+  base: dialogue/base@1.21.2
+  release: dialogue/release@2.17.3
+  python: dialogue/python@3.19.3
+  utils: dialogue/utils@3.17.3
 
 workflows:
   ci:
     jobs:
       - python/pylama:
           name: pylama
-          context: gemfury-download
+          context: code-validator
+      
       - python/isort:
           name: isort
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -100,7 +104,7 @@ jobs:
             - run:
                 name: "install tox"
                 command: |
-                  poetry run pip install -U tox
+                  poetry run pip install -U tox==4.7.0
```
arththebird, line 107: latest version breaks the pipeline

.circleci/config.yml
```diff
@@ -51,15 +56,14 @@ workflows:
     jobs:
       - publish:
           name: publish
-          context: gemfury-download # Needed to build, the upload token is in the CI env vars
+          context: code-validator # Needed to build, the upload token is in the CI env vars
```
edouardlp, line 59: Is the comment still true?
```suggestion
          context: code-validator
```

edouardlp (approved):

.circleci/config.yml
arththebird: yes, because it's public, we upload to pypi and the token is in the CI env vars :)

arththebird: I guess I can just remove it though 🤷

PR #104, opened 2023-10-26
fix: move models to their own modules - DIA-61984

arththebird:
## Description 

Move models to their own module to have better separation of concerns. This change shouldn't cause any issue since we expect people to include the models from the root module like `from fastapi_sqla import Item`.

## Related JIRA issues

* [DIA-61984]

## Validation

All tests still pass ✅ 

### DIFF
--- a/fastapi_sqla/asyncio_support.py
+++ b/fastapi_sqla/asyncio_support.py
@@ -13,7 +13,8 @@ from sqlalchemy.orm.session import sessionmaker
 from sqlalchemy.sql import Select, func, select
 
 from fastapi_sqla import aws_aurora_support, aws_rds_iam_support
-from fastapi_sqla.sqla import Base, Page, T, new_engine
+from fastapi_sqla.models import Page
+from fastapi_sqla.sqla import Base, new_engine
 
 logger = structlog.get_logger(__name__)
 _ASYNC_SESSION_KEY = "fastapi_sqla_async_session"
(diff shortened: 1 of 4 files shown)


fastapi_sqla/asyncio_support.py
```diff
@@ -148,15 +149,15 @@ async def paginate_query(
     limit: int,
     *,
     scalars: bool = True,
-) -> Page[T]:
+) -> Page:
```
arththebird, line 152: It didn't made a lot of sense to return `[T]` here since the TypeVar wasn't bound to anything in that context. It would have made sense if for example we had the signature like:

```
async def paginate_query(
    query: Select[T],
    session: AsyncSession,
    total_items: int,
    offset: int,
    limit: int,
    *,
    scalars: bool = True,
) -> Page[T]:
```

fastapi_sqla/sqla.py
```diff
@@ -293,15 +255,15 @@ def _paginate(
     limit: int,
     *,
     scalars: bool = True,
-) -> Page[T]:
+) -> Page:
```
arththebird, line 258: It didn't made a lot of sense to return `[T]` here since the TypeVar wasn't bound to anything in that context. It would have made sense if for example we had the signature like:

```
def paginate_query(
    query: Select[T],
    session: Session,
    total_items: int,
    offset: int,
    limit: int,
    *,
    scalars: bool = True,
) -> Page[T]:
```

hadrien (approved):

Issue #102, opened 2023-10-17
How to use this package with multiple DBs?


brandonmailhiot:
Could not find documentation about whether this is or isn't supported. It would be great if I could get some guidance on how I can set up my service to use 2+ DBs. I specifically need to read from one Postgres DB and write to two Postgres DBs.

hadrien:
Hi @brandonmailhiot,

It is not something supported.

We [@dialogue](https://github.com/dialoguemd/) are in a need of a read only session to the same db. So this is something we will implement eventually.

hadrien:
@brandonmailhiot FYI we started https://github.com/dialoguemd/fastapi-sqla/pull/105 
Should be released by the end of next week.

arththebird:
This is now supported with version 3.0.0. See the [release notes](https://github.com/dialoguemd/fastapi-sqla/releases/tag/v3.0.0) for more details.

PR #105, opened 2023-10-31
feat: support multiple sessions - DIA-61984

arththebird:
## Description 

Introduce support for multiple sessions running in parallel. This will primarily be used to support read-only sessions using a different user (connection string) and be able to use AWS Aurora read only replicas. To configure a new separate session, you just have to set a `fastapi_sqla__MY_KEY__sqlalchemy_url` environment variable and it will be configured. See the README for more details about how to use it.

Note that the default session is still configured via `sqlalchemy_url` and all the existing functions default to using the default session if the key is not specified.

## Breaking changes
- Remove support for configuring a separate async session with `async_sqlalchemy_url`. The supported way is now to use an async dialect in the default connection string, or use another session with the proper format.
- The `Session`, `AsyncSession`, `Paginate` and `AsyncPaginate` dependencies are now using the `Annotated` form recommended by FastAPI since version 0.95.0. This means that you'll need to change the usage of the dependencies like: `session: Session = Depends()` -> `session: Session`.
- Drop support for python versions lower than `3.9`. This was already the case since we were only testing with those versions, but the support is now explicitly removed from the `pyproject.toml`.
- The library now requires FastAPI `>= 0.95.1`

## Related

* [DIA-61984]
* #102

## Validation

Default session still works ✅  Tested with [this PR](https://github.com/dialoguemd/mattermerde/pull/188)

Integration tests covers using multiple sessions ✅ 


README.md
```diff
@@ -27,7 +27,7 @@ unique `email`:
 
 ```python
 # main.py
-from fastapi import Depends, FastAPI, HTTPException
```
arththebird, line 30: TODO: Document how to use session per key and etc.

fastapi_sqla/__init__.py
```diff
@@ -47,18 +53,41 @@
     has_asyncio_support = False
     asyncio_support_err = str(err)
 
+ENGINE_KEYS_REGEX = re.compile(r"fastapi_sqla__(.+)__.+")
```
arththebird, line 56: TODO: Confirm syntax with team

fastapi_sqla/sqla.py
```diff
@@ -33,27 +33,33 @@
...
-    envvar_prefix = envvar_prefix if envvar_prefix else "sqlalchemy_"
-    lowercase_environ = {
-        k.lower(): v for k, v in os.environ.items() if k.lower() != "sqlalchemy_warn_20"
-    }
+class Base(DeclarativeBase, DeferredReflection):
+    __abstract__ = True
+
+
+def new_engine(*, key: str = "default") -> Engine:
+    envvar_prefix = "sqlalchemy_"
```
arththebird, line 44: TODO: Check if ok to remove ability to override this

fastapi_sqla/sqla.py
```diff
@@ -33,27 +33,33 @@
...
-def new_engine(*, envvar_prefix: Union[str, None] = None) -> Engine:
-    envvar_prefix = envvar_prefix if envvar_prefix else "sqlalchemy_"
-    lowercase_environ = {
-        k.lower(): v for k, v in os.environ.items() if k.lower() != "sqlalchemy_warn_20"
-    }
+class Base(DeclarativeBase, DeferredReflection):
+    __abstract__ = True
+
+
+def new_engine(*, key: str = "default") -> Engine:
```
arththebird, line 43: nit: Make the `default` string a constant and use it everywhere

pyproject.toml
```diff
@@ -40,8 +38,8 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.7"
-fastapi = ">=0.61"
+python = "^3.9"
```
arththebird, line 41: We only test with python 3.9+, so I removed support for lower version. (If we don't test with it, we don't support it)

fastapi_sqla/_pytest_plugin.py
```diff
@@ -121,18 +120,27 @@ def sqla_transaction(sqla_connection):
...
     """Sqla session to use when creating db fixtures.
 
     While it does not write any record in DB, the application will still be able to
     access any record committed with that session.
     """
-    import fastapi_sqla.sqla
-
-    yield fastapi_sqla.sqla._Session(bind=sqla_connection)
+    session = session_factory(bind=sqla_connection)
+    yield session
```
arththebird, line 142: Not absolutely necessary, but it brings it up to par with the async version.

fastapi_sqla/__init__.py
```diff
@@ -47,18 +57,48 @@
...
+    engine_keys = _get_engine_keys()
+    engines = {key: sqla.new_engine(key) for key in engine_keys}
+    for key, engine in engines.items():
+        if not sqla.is_async_dialect(engine):
+            app.add_event_handler("startup", functools.partial(sqla.startup, key=key))
+            app.middleware("http")(
+                functools.partial(sqla.add_session_to_request, key=key)
+            )
+
+        # TODO: Check if we can get rid of it. I think so
```
arththebird, line 93: TODO

fastapi_sqla/sqla.py
```diff
@@ -311,4 +332,5 @@ def paginate(query: DbQuery, scalars=True) -> Page:
         return default_dependency
 
 
-Paginate: PaginateDependency = Pagination()
+Paginate = Annotated[PaginateDependency, Depends(Pagination())]
+Session = Annotated[SqlaSession, Depends(SessionDependency())]
```
arththebird, line 336: TBD, but I might be able to make this non breaking by using this form instead. But I feel that since the upgrade path is very simple, it's a good time to introduce a breaking change. We could also get rid of a couple legacy things.

```suggestion
Session: SqlaSession = SessionDependency()
```

fastapi_sqla/__init__.py
edouardlp: Makes sense to me

fastapi_sqla/async_pagination.py
```diff
@@ -1,28 +1,31 @@
 import math
```
arththebird, line 1: TODO: Tests

fastapi_sqla/async_sqla.py
```diff
@@ -1,88 +1,79 @@
 import os
```
arththebird, line 1: TODO: Tests

fastapi_sqla/pagination.py
```diff
@@ -1,24 +1,24 @@
 import math
```
arththebird, line 1: TODO: Tests

fastapi_sqla/sqla.py
```diff
@@ -2,10 +2,10 @@
 import os
```
arththebird, line 2: TODO: Tests

tests/conftest.py
```diff
@@ -53,29 +53,28 @@ def is_boto3_installed():
...
+        values[
+            f"fastapi_sqla__{async_session_key}__sqlalchemy_url"
+        ] = async_sqlalchemy_url
 
     with patch.dict("os.environ", values=values, clear=True):
         yield values
 
 
-@fixture(scope="session")
-def engine(environ):
```
arththebird, line 72: This was moved to the pytest plugin for better consistency

tests/middleware/conftest.py
```diff
@@ -0,0 +1,89 @@
+from unittest.mock import Mock
```
arththebird, line 1: The middleware tests were refactored to improve coverage for the async version by adopting a similar pattern as the pagination tests. Consider moving this to its own PR.

tests/test_async_sqla.py
```diff
@@ -1,62 +0,0 @@
-from unittest.mock import AsyncMock, patch
-
-from pytest import fixture, mark, raises
-from sqlalchemy import text
-
-pytestmark = [mark.sqlalchemy("1.4"), mark.require_asyncpg]
-
-
-@fixture
-async def startup(environ):
```
arththebird, line 10: The tests related to startup were moved to `test_startup` and the ones related to `open_session` moved to `test_open_session`. Consider moving this to its own PR.

tests/middleware/test_middleware.py
```diff
@@ -0,0 +1,94 @@
...
+    mock_middleware.assert_called_once()
+
+    assert {
+        "event": "http error, rolling back possibly uncommitted changes",
+        "log_level": "warning",
+        "status_code": 404,
+    } not in caplog
+
+
+async def test_async_session_dependency_raises_unknown_key(client):
```
arththebird, line 82:
```suggestion
async def test_session_dependency_raises_unknown_key(client):
```

hadrien:
Just got notified about this hence why I am late to the party. Is there a way to break down these commits in separate PRs?

arththebird:
> Just got notified about this hence why I am late to the party. Is there a way to break down these commits in separate PRs?

No worries at all, I just opened the PR for review, so you're not late :)

I already removed a bunch of refactoring related things and moved it to a couple PRs that were merged and [this one](https://github.com/dialoguemd/fastapi-sqla/pull/107) for test refactor. I don't really see a way to split it further because it's kind of all related to the change from a single session to multiple sessions. 

If you see an opportunity to split it, please let me know and it will be my pleasure to update it. FWIW, there's a lot of file that were changed, but not all of them have huge change, it's just so that it uses key based session instead of using the previous `_Session` (or `_AsyncSession`).

hadrien (commented):
```python
ReadOnlySession = Annotated[SqlaSession, Depends(SessionDependency(key="read_only"))]
```
This is nice. 
1. IIUC, it can be overides in test using something like:
    ```python

    
    def app():
        from ... import ReadOnlySession
        from ...main import app
        app.dependency_override[ReadOnlySession] = Mock()
2. Can we do even simpler? 
    ```python
    from fastqpi_sqla import new_session_dependency
    
    ReadOnlySession = new_session_dependency("read-only")
    ```

hadrien:
3. 
    ```python
    ReadOnlySession = Annotated[SqlaSession, Depends(SessionDependency(key="read_only"))]
    ```
    Not sure if this is the case but does this fail only at startup rather than at import?

arththebird:
> 3. ```python
>    ReadOnlySession = Annotated[SqlaSession, Depends(SessionDependency(key="read_only"))]
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    Not sure if this is the case but does this fail only at startup rather than at import?

It keeps the same behaviour as before where it only fails at runtime instead of at import. At startup, if you configured a Session with an invalid URL or the db is not accessible, it will fail like before, but for the dependency itself, it only fails at runtime because it's like if you didn't called the `fastapi_sqla.setup` function that adds the middleware that instantiate the session for the request.

arththebird:
> ```python
> ReadOnlySession = Annotated[SqlaSession, Depends(SessionDependency(key="read_only"))]
> ```
> 
> This is nice.
> 
> 1. IIUC, it can be overides in test using something like, right?
>    ```python
>    def app():
>        from ... import ReadOnlySession
>        from ...main import app
>        app.dependency_override[ReadOnlySession] = Mock()
>    ```
> 2. Can we do even simpler?
>    ```python
>    from fastapi_sqla import new_session_dependency
>    
>    ReadOnlySession = new_session_dependency("read-only")
>    ```

Good catch for the dependency override in tests! I thought it would work OOTB, but it doesn't. [More details here as to why](https://github.com/tiangolo/fastapi/discussions/9452).

So to have it work, we would need to do something like:
```python
read_only_session_dep = SessionDependency(key="read_only")
ReadOnlySession = Annotated[SqlaSession, Depends(read_only_session_dep)]

def app():
    from ... import read_only_session_dep
    from ...main import app
    app.dependency_override[read_only_session_dep] = Mock()
```

For doing simpler, maybe we could, but it would need more investigation to see whether the type hints are understood by the editors when returning an Annotated type from a function. Also, because of the above, it would only work when we don't want to override it in tests. It definitively is a good suggestion, but I suggest we look into it as a separate story because it's a NTH in my opinion :)

fastapi_sqla/async_sqla.py
```diff
@@ -154,3 +139,38 @@ async def get_users(session: fastapi_sqla.AsyncSession = Depends()):
...
+        except KeyError:
+            logger.exception(
+                f"No async session with key '{self.key}' found in request, "
+                "please ensure you've setup fastapi_sqla.",
+                session_key=self.key,
+            )
+            raise
+
+
+default_async_session_dep = AsyncSessionDependency()
```
arththebird, line 175: Addresses [this comment](https://github.com/dialoguemd/fastapi-sqla/pull/105#issuecomment-1795317685) for easier overrides

fastapi_sqla/sqla.py
```diff
@@ -155,3 +145,36 @@ def get_users(session: fastapi_sqla.Session = Depends()):
...
+        except KeyError:
+            logger.exception(
+                f"No session with key '{self.key}' found in request, "
+                "please ensure you've setup fastapi_sqla.",
+                session_key=self.key,
+            )
+            raise
+
+
+default_session_dep = SessionDependency()
```
arththebird, line 179: Addresses [this comment](https://github.com/dialoguemd/fastapi-sqla/pull/105#issuecomment-1795317685) for easier overrides

fastapi_sqla/sqla.py
```diff
@@ -119,11 +108,12 @@ async def add_session_to_request(request: Request, call_next):
...
 
         @app.get("/users")
-        def get_users(session: fastapi_sqla.Session = Depends()):
+        def get_users(session: fastapi_sqla.Session):
             return session.execute(...) # use your session here
     """
-    async with contextmanager_in_threadpool(open_session()) as session:
-        request.scope[_SESSION_KEY] = session
+    async with contextmanager_in_threadpool(open_session(key)) as session:
+        # TODO: This should use request.state
```
edouardlp, line 115: Is this for a future PR?

arththebird: Fixed ✅

edouardlp: Broken you mean 😅 ?

arththebird: oops, I talked too fast 😂

arththebird: Alright, fixed for real this time 😄 That's what happens when I do thee things at the same time 😅

edouardlp (approved):
Very nice!

hadrien:
Bien joué. We have never ever overridden the session dependency anyway so that was a silly comment.

README.md
```diff
@@ -153,36 +168,87 @@ class Entity(Base):
...
 
+# Preferred
+
+read_only_session_dep = SessionDependency(key="read_only")
+ReadOnlySession = Annotated[SqlaSession, Depends(read_only_session_dep)]
+
+async_read_only_session_dep = AsyncSessionDependency(key="read_only")
+AsyncReadOnlySession = Annotated[
+    SqlaAsyncSession, Depends(async_read_only_session_dep)
+]
```
hadrien, line 216: I'd get back to how it was documented without separating the dep from annotation. Seems cleaner to me. And sorry for suggesting it had to be overridden in tests... YAGNI

hadrien (approved):

PR #106, opened 2023-11-01
feat: split the main modules into multiple modules - DIA-61984

arththebird:
## Description 

Split the main `sqla` and `asyncio_support` modules into multiple smaller modules with their own purpose. This should help properly separate concerns instead of having everything bundled together. This change shouldn't cause any issue since we expect people to import the classes and objects from the root module like `from fastapi_sqla import Session`.

#### Additional details
- Pagination code now lives in `pagination.py` and `async_pagination.py`
- The `Base` model is now part of `models.py`
- `asyncio_support.py` was renamed to `async_sqla.py` since it's not containing all the async code anymore, and because it makes more sense with the addition of `async_pagination.py`. It shows in a better way that it's the async version of the `sqla` module.
- The `setup` function was moved to `base.py` so that the `__init__.py` file only contains imports of what we want to expose from the root module. This uses a similar design as the pydantic or fastapi packages.

## Related JIRA issues

* [DIA-61984]

## Validation

No code change, only moving things around. Tests still pass with minimal changes.


hadrien:
Thanks! well done.

hadrien (approved):

PR #107, opened 2023-11-03
fix: async middleware is equivalent to sync middleware - DIA-61984

arththebird:
## Description 

- Update async middleware and context manager to be equivalent to the sync version.
- Refactor async tests to be equivalent to sync version.
- Update pytest plugin to be consistent between sync and async version.

## Related PRs

Extracted from #105 

## Related JIRA issues

* [DIA-61984]

## Validation

New behaviour is thoroughly tested with unit and integration tests ✅ 


tests/conftest.py
```diff
@@ -88,6 +80,7 @@ def tear_down(environ):
     # reload fastapi_sqla to clear sqla deferred reflection mapping stored in Base
     importlib.reload(fastapi_sqla.models)
     importlib.reload(fastapi_sqla.sqla)
+    importlib.reload(fastapi_sqla.async_sqla)
```
arththebird, line 83: Forgotten in previous refactor

tests/middleware/conftest.py
```diff
@@ -0,0 +1,83 @@
+from unittest.mock import Mock
```
arththebird, line 1: The middleware tests were refactored to improve coverage for the async version by adopting a similar pattern as the pagination tests.

tests/test_middleware.py
```diff
@@ -1,191 +0,0 @@
-from unittest.mock import Mock, patch
```
arththebird, line 1: The tests related to startup were moved to test_startup and the ones related to open_session moved to test_open_session.

tests/test_startup.py
```diff
@@ -6,17 +6,6 @@
 from pytest import fixture, mark, raises
 from sqlalchemy import text
 
-pytestmark = mark.usefixtures("patch_engine_from_config", "patch_new_engine")
-
-
-@fixture(params=[True, False])
-def case_sensitive_environ(environ, request):
```
arththebird, line 13: It was causing issue with the `PYTEST_CURRENT_TEST` env var, so moved to its own test

fastapi_sqla/_pytest_plugin.py
```diff
@@ -92,19 +96,19 @@ def sqla_modules():
...
 
 
 @fixture
-def patch_engine_from_config(request, db_url, sqla_connection, sqla_transaction):
+def patch_engine_from_config(request, sqla_connection, sqla_transaction):
     """So that all DB operations are never written to db for real."""
     from fastapi_sqla.sqla import _Session
 
-    if "dont_patch_engines" in request.keywords:
+    if "dont_patch_engines" in request.keywords:  # pragma: no cover
```
arththebird, line 111: We weren't actually testing it, and it was flagged as not covered after I cleaned up `test_startup`. I don't think it should be that PR responsibility to add coverage for this case in this fixture

hadrien: 🤔 weird that it was flagged as not covered as that marker is used in tests.

hadrien: (Anyway, agree it is not the responsibility of current PR to cover this)

edouardlp (approved):

PR #114, opened 2024-01-22
feat: Add support for SQLModel - DIA-65344

hadrien:
## Problem

* FastAPI-SQLA is not compatible with [SQLModel](https://sqlmodel.tiangolo.com/).
* Would be nice if fastapi-sqla can capture a bit of the SQLModel audience.

## Solution

* Instantiate an SQLModel session if sqlmodel can be imported.
* SQLAlchemy's `sessionmaker` expect a `class_` argument and because SQLModel's Session inherits from SQLAlchemy's Session, it just works.

Also:
* Move `fastapi_sqla.models.Base` to `fastapi_sqla.sqla`: It is not a model but the SQLAlchemy declarative base.
* To allow testing with and without sqlmodel and avoid busting the circleci test matrix, remove testing with pydantic 2.0 and 2.1 to only test with pydantic <2 and >=2.

## Related

* [DIA-65344]
* [SQLModel]

## Validation

* Added a test module `test_sqlmodel`

[DIA-65344]: https://dialoguemd.atlassian.net/browse/DIA-65344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SQLModel]: https://sqlmodel.tiangolo.com


fastapi_sqla/sqla.py
```diff
@@ -50,7 +70,12 @@ def startup(key: str = _DEFAULT_SESSION_KEY):
         raise
 
     Base.prepare(engine)
-    _session_factories[key] = sessionmaker(bind=engine)
+
+    if has_sqlmodel:
+        _session_factories[key] = sessionmaker(bind=engine, class_=SQLModelSession)
```
vicrep, line 75: Nit, but since we determine whether SQLModel is supported at import time, would it not be a little less indirected to assign the appropriate session class at same time as import?

```py

try:
    from sqlmodel import Session as SessionCls
except ImportError:
    from sqlalchemy.orm.session import Session as SessionCls
```

And IIUC this should still work with typings since type checkers actually import + interpret modules.

vicrep (approved):
LGTM! Really cool how SQLModel decided to extend SQLA session, makes it such a transparent addition!

fastapi_sqla/sqla.py
hadrien: Done, I kept using `SqlaSession` to avoid breaking change (even though SQLASession would be better)

fastapi_sqla/sqla.py
```diff
@@ -146,11 +165,14 @@ def get_users(session: fastapi_sqla.Session):
     return response
 
 
-class SessionDependency:
+S = TypeVar("S", bound=SqlaSession)
+
+
+class SessionDependency(Generic[S]):
```
arththebird, line 171: Sorry, I'm very late to the party. I'm curious how come this was needed since `SqlaSession` is overloaded depending on when SQLModel is imported or not 🤔 

Very nice feature btw! 🙌

hadrien: I changed how it was done and forgot to remove this. good catch, will cleanup

hadrien: (Before, there were 2 session types `SqlaSession` & `SQLModelSession`)

arththebird: Got it, thanks for the explanation!

hadrien: fixed in #121

PR #132, opened 2024-06-10
feat: Add support for FastAPI Lifespan DIA-69117

vtremblay:
## Problem

FastAPI SQLA Cannot be used with the FastAPI Lifespan, only with the deprecated startup/shutdown events.

## Solution

Split the setup function into two smaller ones; the first one should be called in the Lifespan function and the second one is used to register the middlewares (which cannot be done in the lifespan)

Note that the previous `setup` function is still there for backward compatibility

## Validation

Tested in ER:
1) The app boot correctly using a Lifespan instead of the events
2) The test are all passing


fastapi_sqla/base.py
```diff
@@ -20,6 +20,31 @@
 _ENGINE_KEYS_REGEX = re.compile(r"fastapi_sqla__(?!_)(.+)(?<!_)__(?!_).+")
 
 
+async def engines_statup():
```
arththebird, line 23: nit: typo
```suggestion
async def engines_startup():
```

Also, should we just make it `startup` so that it's our entry point to anything that should happen at startup?

```suggestion
async def startup():
```

vtremblay: good point for startup

README.md
```diff
@@ -27,15 +27,22 @@ unique `email`:
 
 ```python
 # main.py
+from contextlib import asynccontextmanager
 from fastapi import FastAPI, HTTPException
-from fastapi_sqla import Base, Item, Page, Paginate, Session, setup
+from fastapi_sqla import Base, engines_statup, Item, Page, Paginate, Session, setup_middlewares
```
arththebird, line 32: Following the change to startup, let's update the doc :)

fastapi_sqla/base.py
```diff
@@ -20,6 +20,31 @@
...
+            app.middleware("http")(
+                functools.partial(sqla.add_session_to_request, key=key)
+            )
+        else:
+            app.middleware("http")(
+                functools.partial(async_sqla.add_session_to_request, key=key)
+            )
+
+
+# TODO: Remove this function once all FastAPI apps stops using startup/shutdown events
```
arththebird: Since this is a OOS lib, should we use the [deprecated](https://github.com/tantale/deprecated) lib and add a deprecated decorator here to tell consumers to upgrade to the new startup, because it'll be deprecated in v4?

README.md
```diff
@@ -140,7 +147,23 @@ And define the environment variable `sqlalchemy_url` with `postgres+asyncpg` sch
...
+async def lifespan(app: FastAPI):
+    await fastapi_sqla.startup(app)
+    yield
+
+
+app = FastAPI(lifespan=lifespan)
+fastapi_sqla.setup_middlewares(app)
+```
+
+## Setup the app using startup/shutdown events (legacy):
```
arththebird, line 166: nit
```suggestion
## Setup the app using startup/shutdown events (deprecated):
```

fastapi_sqla/__init__.py
```diff
@@ -22,6 +22,8 @@
     "SqlaSession",
     "open_session",
     "setup",
+    "setup_middlewares",
```
arththebird, line 25: nit: let's add it to the `test_import_sync_api` test

fastapi_sqla/base.py
vtremblay: done

pyproject.toml
```diff
@@ -60,6 +60,7 @@ boto3 = {version = "^1.24.74", optional = true}
 greenlet = {version = "^3.0.3", optional = true}
 mypy = {version = "^1.0.0", extras = ["tests"], optional = true}
 sqlmodel = { version = "^0.0.14", optional = true }
+deprecated = "^1.2.14"
```
arththebird, line 63: nit: since this is a real deps, let move it to the top

tests/test_import.py
```diff
@@ -28,4 +29,5 @@ def test_import_async_api():
         AsyncSessionDependency,
         SqlaAsyncSession,
         open_async_session,
+        startup,
```
arththebird, line 32: I think this should go to the test above since it's required for both sync and async sessions :)

vtremblay: yeah, I was not sure where to put it

fastapi_sqla/base.py
```diff
@@ -20,6 +20,31 @@
 _ENGINE_KEYS_REGEX = re.compile(r"fastapi_sqla__(?!_)(.+)(?<!_)__(?!_).+")
 
 
+async def startup():
```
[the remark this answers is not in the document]
vtremblay: fixed

tests/test_setup.py
```diff
@@ -1,10 +1,22 @@
-from unittest.mock import Mock, patch
```
edouardlp, line 1: It's a bit confusing for the new `startup` function to be tested in a file called `test_setup`, when `setup` is deprecated. Should we rename the file to `test_startup.py` and rename the previous `test_startup.py` to `test_sqla_startup.py`.  Alternatively we should merge both, but we should make it clear when were testing the wrapper vs the specific implementation.

fastapi_sqla/base.py
```diff
@@ -20,6 +21,33 @@
...
+    engine_keys = _get_engine_keys()
+    engines = {key: sqla.new_engine(key) for key in engine_keys}
+    for key, engine in engines.items():
+        if not _is_async_dialect(engine):
+            sqla.startup(key=key)
+        else:
+            await async_sqla.startup(key=key)
```
edouardlp, line 31: Should we add tests for the async cases?

vtremblay: hum, that's a lie

```
@mark.sqlalchemy("1.4")
@mark.require_asyncpg
async def test_startup_with_sync_and_async_sqlalchemy_url(
    async_session_key, sqla_startup_mock, async_sqla_startup_mock
):
    from fastapi_sqla.base import startup
    from fastapi_sqla.sqla import _DEFAULT_SESSION_KEY

    await startup()

    assert async_sqla_startup_mock.call_count == 1
    async_sqla_startup_mock.assert_awaited_once_with(key=async_session_key)

    assert sqla_startup_mock.call_count == 1
    sqla_startup_mock.assert_called_once_with(key=_DEFAULT_SESSION_KEY)
    ```

tests/test_setup.py
vtremblay: right. I have renamed `test_startup.py` to `test_sqla_startup.py` and split `test_setup.py` into 3 files.

edouardlp (approved):

hadrien:
Why not change `setup` to use lifespan rather than having library users to do so?

hadrien:
Also, why changing `setup` to `setup_middlewares`? is that necessary? Whenever we'll want to setup `something_else`, will we add a `setup_something_else` 
Can't we just keep `setup`?

PR #138, opened 2024-06-17
fix: standardize deps and renovate config - DIA-70164

edouardlp:
## Problem

To work around limitations of tox and poetry, we weren't using dev dependencies and extras properly. This creates issues when renovate determines the semantic commit type, leading to incorrect versioning of the library.

## Solution

- We use an arguably uglier tox configuration that has become our new standard that allows us to separate main and dev dependencies.
- We leverage our library renovate preset which ensures dependency ranges are extended rather than replaced and which further tunes semantic commit type.

## Validation

✅ poetry run tox -e poetry run tox -e sqlalchemy(version) performs the correct installs (beyond `poetry install`)

| version | commands                         |
|--------------------|----------------------------------|
| 1.3                | pip install sqlalchemy==1.3      |
| 1.4                | pip install sqlalchemy==1.4.52   |
| 2.0                | None                             |
| 2.0-sqlmodel       | poetry install --extras sqlmodel |

✅ poetry run tox -e sqlalchemyXX-(packages) installs the proper packages at the end

Note that this is clearly inefficient, but it's one of the tradeoffs of this approach, and in practice only adds a few ms to the test execution

| packages                                   | extras                                                                           |
|--------------------------------------------|----------------------------------------------------------------------------------|
| sqlmodel-pydantic2-noasyncpg-noaws_rds_iam | "sqlmodel"                                                                         |
| sqlmodel-pydantic2-asyncpg-noaws_rds_iam   | "sqlmodel", "asyncpg", "sqlmodel asyncpg"                                        |
| sqlmodel-pydantic2-asyncpg-aws_rds_iam     | "sqlmodel", "aws_rds_iam", "asyncpg aws_rds_iam", "sqlmodel asyncpg aws_rds_iam" |
| sqlmodel-pydantic2-noasyncpg-aws_rds_iam   | "sqlmodel", "aws_rds_iam", "sqlmodel aws_rds_iam"                                |
| pydantic2-noasyncpg-noaws_rds_iam | none                                                                         |
| pydantic2-asyncpg-noaws_rds_iam   |  "asyncpg"                                        |
| pydantic2-asyncpg-aws_rds_iam     |  "aws_rds_iam", "asyncpg aws_rds_iam" |
| pydantic2-noasyncpg-aws_rds_iam   | "aws_rds_iam"                                |

✅ poetry run tox -e sqlalchemyXX-pydantic1-YY instals pydantic 1 at the end with `pip install pydantic==1.10.16`

✅ Coverage remains 100% which means we test all the conditional tests


pyproject.toml
```diff
@@ -193,21 +179,22 @@ envlist = sqlalchemy{ 1.3, 1.4, 2.0, 2.0-sqlmodel }-{ asyncpg, noasyncpg }-{aws_
...
+    poetry install
+    sqlmodel: poetry install --extras "sqlmodel"
+    sqlmodel-noasyncpg-aws_rds_iam: poetry install --extras "sqlmodel aws_rds_iam"
+    sqlmodel-asyncpg-noaws_rds_iam: poetry install --extras "sqlmodel asyncpg"
+    sqlmodel-asyncpg-aws_rds_iam: poetry install --extras "sqlmodel asyncpg aws_rds_iam"
+    noasyncpg-aws_rds_iam: poetry install --extras "aws_rds_iam"
+    asyncpg-noaws_rds_iam: poetry install --extras "asyncpg"
+    asyncpg-aws_rds_iam: poetry install --extras "asyncpg aws_rds_iam"
+    sqlalchemy1.3: pip install sqlalchemy==1.3
+    sqlalchemy1.4: pip install sqlalchemy==1.4.52
```
edouardlp, line 194: 1.4 was failing.

pyproject.toml
```diff
@@ -38,52 +38,38 @@ classifiers = [
...
 
 [tool.poetry.dependencies]
-python = "^3.9"
-fastapi = ">=0.95.1"
-pydantic = ">=1"
-sqlalchemy = ">=1.3"
-structlog = ">=20"
-deprecated = ">=1.2"
+python = ">=3.9,<3.13"
+fastapi = ">=0.95.1,<0.112"
```
edouardlp, line 42: As discussed in similar PRs. This is required given that semantic versioning lets fastapi introduce breaking changes until 1.0. It has the drawback that we have to extend the range before any user can update.

README.md
```diff
@@ -600,7 +602,7 @@ If your project uses [SQLModel], then `Session` dependency is an SQLModel sessio
 # Pytest fixtures
 
 This library provides a set of utility fixtures, through its PyTest plugin, which is
-automatically installed with the library.
+automatically installed with the library. Using the plugin requires the `pytest_plugin` extra.
```
arththebird, line 605: I guess we don't consider this a breaking change because the extra is there, and most people would actually have this installed directly instead of via this lib?

pyproject.toml
```diff
@@ -193,21 +179,22 @@ envlist = sqlalchemy{ 1.3, 1.4, 2.0, 2.0-sqlmodel }-{ asyncpg, noasyncpg }-{aws_
 
 [testenv]
 passenv = CI
```
arththebird, line 181: is this still necessary?

README.md
edouardlp: It's already the case that this is required as I found out in [this PR](https://github.com/dialoguemd/guestlist/pull/298). Removing `alembic` from a repo using the pytest plugin fails. So offering this is not a breaking change but documenting the current state.

pyproject.toml
arththebird: ah yeah, I think it's used in the pytest plugin

edouardlp: Yes, the CI injects the postgres variables that are needed to connect to the DB.

edouardlp: Yeah [this](https://github.com/dialoguemd/fastapi-sqla/blob/master/fastapi_sqla/_pytest_plugin.py#L31)

pyproject.toml
```diff
@@ -193,21 +179,22 @@ envlist = sqlalchemy{ 1.3, 1.4, 2.0, 2.0-sqlmodel }-{ asyncpg, noasyncpg }-{aws_
...
-    tests
-
-commands = pytest -vv --cov={envsitepackagesdir}/fastapi_sqla --cov-report xml --cov-report html --junitxml=test-reports/pytest/junit.xml
+skip_install = true
+allowlist_externals = poetry
+commands_pre =
+    poetry install
+    sqlmodel: poetry install --extras "sqlmodel"
+    aws_rds_iam: poetry install --extras "aws_rds_iam"
+    asyncpg-noaws_rds_iam: poetry install --extras "asyncpg"
```
arththebird, line 188: we specify it here because else it would install asyncpg twice in `asyncpg-aws_rds_iam`?

arththebird (approved):

pyproject.toml
edouardlp: I assume your question is : why do we have `asyncpg-noaws_rds_iam` instead of just `asyncpg`? Good catch. It does make it more efficient, but at the cost of a more complex config, and is not consistent with how I've approached the other dependencies.

edouardlp: addressed in https://github.com/dialoguemd/fastapi-sqla/pull/138/commits/05806c24041d7ed620b2851c8e0f59e731bc2647

pyproject.toml
```diff
@@ -193,21 +179,22 @@ envlist = sqlalchemy{ 1.3, 1.4, 2.0, 2.0-sqlmodel }-{ asyncpg, noasyncpg }-{aws_
...
+skip_install = true
+allowlist_externals = poetry
+commands_pre =
+    poetry install
+    sqlmodel: poetry install --extras "sqlmodel"
+    aws_rds_iam: poetry install --extras "aws_rds_iam"
+    asyncpg: poetry install --extras "asyncpg"
+    asyncpg-aws_rds_iam: poetry install --extras "asyncpg aws_rds_iam"
+    sqlmodel-noasyncpg-aws_rds_iam: poetry install --extras "sqlmodel aws_rds_iam"
+    sqlmodel-asyncpg-noaws_rds_iam: poetry install --extras "sqlmodel asyncpg"
```
arththebird, line 191: a bit similar to [this comment](https://github.com/dialoguemd/fastapi-sqla/pull/138#discussion_r1644880497), should we do
```suggestion
    sqlmodel-aws_rds_iam: poetry install --extras "sqlmodel aws_rds_iam"
    sqlmodel-asyncpg: poetry install --extras "sqlmodel asyncpg"
```

edouardlp: I'll try but I don't think this will work because of my mental model of how the badly documented `commands_pre` works.

edouardlp: It does work 😅.

edouardlp: https://github.com/dialoguemd/fastapi-sqla/pull/138/commits/0679b79d1a4e497de1755218fafd9a8fdf8de8e8

PR #142, opened 2024-06-20
fix: use lifespan event to startup

hadrien:
## Problem

* Having to setup fastapi lifespan to use the library is less simple that not.
* Deprecating the main _setup_ entry point of library is wrong IMO

## Solution

* Move back to `fastapi_sqla.setup` as the main setup entry point.
* `fastapi_sqla.setup` runs startups when lifespan event is being received.

## Related

* #132 
* [DIA-69117]

## Validation

[DIA-69117]: https://dialoguemd.atlassian.net/browse/DIA-69117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hadrien:
Unfortunately, I cannot stay in front of a screen for more than 30minutes so, I will not be able to followup on this.

hadrien:
Closing as I cannot follow up

PR #145, opened 2024-06-21
feat: use pure ASGI middlewares

alixlahuec:
## Problem

When using multiple sessions, we've seen sporadic issues coming from the injected middlewares where the response body is missing, causing errors due to a mismatch with the promised `Content-Length`. We've also observed a significant increase in endpoint latency.

Besides, the middleware class we're using, [`BaseHTTPMiddleware`](https://www.starlette.io/middleware/#basehttpmiddleware), is no longer the recommended way of writing middlewares for Starlette. It's known to have functional issues and does not scale well (see [discussion on this FastAPI issue](https://github.com/tiangolo/fastapi/discussions/6985#discussioncomment-8989601), [discussion on this Starlette issue](https://github.com/encode/starlette/issues/1634#issuecomment-1124806406), and [this performance benchmark](https://kisspeter.github.io/fastapi-performance-optimization/middleware.html)).

## Solution

Refactor the middleware used to inject sessions into the request, to be [pure ASGI middleware](https://www.starlette.io/middleware/#pure-asgi-middleware).

## Related

* [DIA-70887]
* https://github.com/dialoguemd/coredata/pull/1856

## Validation

✅  Tests are passing
✅  ER smoke tests with 2 sessions - calls are successful, both middlewares are present and executed ([sample request - only the default session is used](https://app.datadoghq.com/logs?query=%40request_headers.x-datadog-trace-id%3A16365759639994361819&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=paused&storage=hot&stream_sort=desc&view=spans&viz=stream&from_ts=1718998428120&to_ts=1718998434278&live=false), [sample request - both sessions are used](https://app.datadoghq.com/logs?query=%40request_headers.x-datadog-trace-id%3A16918563406151202983&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=paused&storage=hot&stream_sort=desc&view=spans&viz=stream&from_ts=1718998620000&to_ts=1718998800000&live=false))

[DIA-61986]: https://dialoguemd.atlassian.net/browse/DIA-61986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DIA-70887]: https://dialoguemd.atlassian.net/browse/DIA-70887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fastapi_sqla/async_sqla.py
+++ b/fastapi_sqla/async_sqla.py
@@ -9,6 +9,7 @@ from sqlalchemy import text
 from sqlalchemy.ext.asyncio import AsyncEngine
 from sqlalchemy.ext.asyncio import AsyncSession as SqlaAsyncSession
 from sqlalchemy.orm.session import sessionmaker
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 from fastapi_sqla import aws_aurora_support, aws_rds_iam_support
 from fastapi_sqla.sqla import _DEFAULT_SESSION_KEY, Base, new_engine
@@ -88,9 +89,7 @@ async def open_session(
         await session.close()
 
 
-async def add_session_to_request(
-    request: Request, call_next, key: str = _DEFAULT_SESSION_KEY
-):
+class AsyncSessionMiddleware:
     """Middleware which injects a new sqla async session into every request.
 
     Handles creation of session, as well as commit, rollback, and closing of session.
@@ -108,36 +107,58 @@ async def add_session_to_request(
         async def get_users(session: fastapi_sqla.AsyncSession):
             return await session.execute(...) # use your session here
     """
-    async with open_session(key) as session:
-        setattr(request.state, f"{_ASYNC_REQUEST_SESSION_KEY}_{key}", session)
-        response = await call_next(request)
+    def __init__(self, app: ASGIApp, key: str = _DEFAULT_SESSION_KEY) -> None:
+        self.app = app
+        self.key = key
 
-        is_dirty = bool(session.dirty or session.deleted or session.new)
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            return await self.app(scope, receive, send)
 
-        # try to commit after response, so that we can return a proper 500 response
-        # and not raise a true internal server error
-        if response.status_code < 400:
-            try:
-                await session.commit()
-            except Exception:
-                logger.exception("commit failed, returning http error")
-                response = PlainTextResponse(
-                    content="Internal Server Error", status_code=500
-                )
+        async with open_session(self.key) as session:
+            request = Request(scope=scope, receive=receive, send=send)
+            setattr(request.state, f"{_ASYNC_REQUEST_SESSION_KEY}_{self.key}", session)
 
-        if response.status_code >= 400:
-            # If ever a route handler returns an http exception, we do not want the
-            # session opened by current context manager to commit anything in db.
-            if is_dirty:
-                # optimistically only log if there were uncommitted changes
-                logger.warning(
-                    "http error, rolling back possibly uncommitted changes",
-                    status_code=response.status_code,
-                )
-            # since this is no-op if session is not dirty, we can always call it
-            await session.rollback()
+            async def send_wrapper(message: Message) -> None:
+                if message["type"] == "http.response.start":
+                    has_responded = False
+                    status_code = message["status"]
 
-    return response
+                    is_dirty = bool(session.dirty or session.deleted or session.new)
+
+                    # try to commit after response, so that we can return a proper 500
+                    # and not raise a true internal server error
+                    if status_code < 400:
+                        try:
+                            await session.commit()
+                        except Exception:
+                            logger.exception("commit failed, returning http error")
+                            status_code = 500
+                            response = PlainTextResponse(
+                                content="Internal Server Error", status_code=500
+                            )
+                            await response(scope, receive, send)
+                            has_responded = True
+
+                    if status_code >= 400:
+                        # If ever a route handler returns an http exception,
+                        # we do not want the current session to commit anything in db.
+                        if is_dirty:
+                            # optimistically only log if there were uncommitted changes
+                            logger.warning(
+                                "http error, rolling back possibly uncommitted changes",
+                                status_code=status_code,
+                            )
+                        # since this is no-op if the session is not dirty,
+                        # we can always call it
+                        await session.rollback()
+
+                    if has_responded:
+                        return
+
+                await send(message)
+
+            await self.app(scope, receive, send_wrapper)
 
 
 class AsyncSessionDependency:
--- a/fastapi_sqla/async_sqla.py
+++ b/fastapi_sqla/async_sqla.py
@@ -3,7 +3,7 @@ from contextlib import asynccontextmanager
 from typing import Annotated
 
 import structlog
-from fastapi import Depends, Request
+from fastapi import Depends, Request, Response
 from fastapi.responses import PlainTextResponse
 from sqlalchemy import text
 from sqlalchemy.ext.asyncio import AsyncEngine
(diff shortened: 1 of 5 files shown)


fastapi_sqla/async_sqla.py
```diff
@@ -108,36 +107,59 @@ async def add_session_to_request(
...
+                        try:
+                            await session.commit()
+                        except Exception:
+                            logger.exception("commit failed, returning http error")
+                            status_code = 500
+                            response = PlainTextResponse(
+                                content="Internal Server Error", status_code=500
+                            )
+                            await response(scope, receive, send)
+                            has_responded = True
```
alixlahuec, line 142: 📝  Not returning directly here because we want to execute the block below with the rollback

fastapi_sqla/async_sqla.py
```diff
@@ -108,36 +107,59 @@ async def add_session_to_request(
...
+                    if status_code < 400:
+                        try:
+                            await session.commit()
+                        except Exception:
+                            logger.exception("commit failed, returning http error")
+                            status_code = 500
+                            response = PlainTextResponse(
+                                content="Internal Server Error", status_code=500
+                            )
+                            await response(scope, receive, send)
```
alixlahuec, line 141: 📝  Sending the response here means that the logs below are timestamped after the request, which isn't great. We could prepare the response here, handle log/rollback next, then have a conditional block like this:

```python
if response:
   await response(scope, receive, send)
else:
   await send(message)
```

But I'm not sure what we think is preferable/more readable

fastapi_sqla/async_sqla.py
```diff
@@ -108,36 +107,59 @@ async def add_session_to_request(
...
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            return await self.app(scope, receive, send)
+
+        async with open_session(self.key) as session:
+            request = Request(scope=scope, receive=receive, send=send)
+            setattr(request.state, f"{_ASYNC_REQUEST_SESSION_KEY}_{self.key}", session)
+
+            async def send_wrapper(message: Message) -> None:
+                if message["type"] == "http.response.start":
```
arththebird, line 124: nit: to reduce indentation, should we invert the condition here?
```suggestion
                if message["type"] != "http.response.start":
                  await send(message)
```

fastapi_sqla/async_sqla.py
arththebird: It might be better to prepare and send later. It would fix the timestamp of logs issue and also remove the need for the `has_responded` flag IIUC

fastapi_sqla/sqla.py
```diff
@@ -130,39 +129,62 @@ async def add_session_to_request(
...
+
+    def __init__(self, app: ASGIApp, key: str = _DEFAULT_SESSION_KEY) -> None:
+        self.app = app
+        self.key = key
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            return await self.app(scope, receive, send)
+
+        logger.debug(f"executing fastapi-sqla middleware - {self.key}")
```
arththebird, line 141: Should we remove this?
```suggestion
```

fastapi_sqla/sqla.py
```diff
@@ -130,39 +129,62 @@ async def add_session_to_request(
...
+
+        async with contextmanager_in_threadpool(open_session(self.key)) as session:
+            request = Request(scope=scope, receive=receive, send=send)
+            setattr(request.state, f"{_REQUEST_SESSION_KEY}_{self.key}", session)
+
+            async def send_wrapper(message: Message) -> None:
+                if message["type"] == "http.response.start":
```
arththebird, line 148: Same as for the async middleware wrt to inverting the condition

tests/test_base_setup_middlewares.py
```diff
@@ -79,12 +78,6 @@ def test_setup_middlewares_with_async_default_sqlalchemy_url(async_sqlalchemy_ur
...
-    assert (
-        app.middleware.return_value.call_args.args[0].func
-        == async_sqla.add_session_to_request
-    )
-    assert app.middleware.return_value.call_args.args[0].keywords == {
-        "key": _DEFAULT_SESSION_KEY
-    }
+    app.add_middleware.assert_called_once()
+    assert app.add_middleware.call_args.args[0] == async_sqla.AsyncSessionMiddleware
+    assert app.add_middleware.call_args.kwargs["key"] == _DEFAULT_SESSION_KEY
```
arththebird, line 83: nit: could be simplified with `assert_called_once_with`

tests/test_setup.py
```diff
@@ -130,15 +128,9 @@ def test_setup_with_async_default_sqlalchemy_url(async_sqlalchemy_url):
...
-    app.middleware.assert_called_once_with("http")
-    app.middleware.return_value.assert_called_once()
-    assert (
-        app.middleware.return_value.call_args.args[0].func
-        == async_sqla.add_session_to_request
-    )
-    assert app.middleware.return_value.call_args.args[0].keywords == {
-        "key": _DEFAULT_SESSION_KEY
-    }
+    app.add_middleware.assert_called_once()
```
arththebird, line 131: nit: same as https://github.com/dialoguemd/fastapi-sqla/pull/145/files#r1653167784

fastapi_sqla/sqla.py
alixlahuec: Ah yes, I ended up making a separate branch for validation and forgot to clean this up - removing!

fastapi_sqla/async_sqla.py
alixlahuec: We can, it's a bit ugly to have `await send(message)` twice but it'll keep things consistent with the check on scope type:

```python
if message["type"] != "http.response.start":
   return await send(message)

# ... check status code, handle rollback, etc

if response:
   return await response(scope, receive, send)

return await send(message)
```

arththebird: I would argue that it's not ugly and actually makes the code more legible, because past the return condition, you only have to care about the use case we want to handle to understand the code 😄

alixlahuec: Fair 😅  So many conditional blocks haha

arththebird (approved):
Very nice! 🙌

alixlahuec:
✅  Did a final smoke test with ER, with both sessions & read/writes 🚀

hadrien:
FWIW, when I started this lib, I initially wrote a fastapi dependency to handle the sqla session. 

But there was a bug/flaw in fastapi (https://github.com/fastapi/fastapi/issues/869).

Therefore,  @vicrep and I went with a middleware instead.
 
Today, that fastapi bug/flaw has been fixed: I strongly recommend ditching the middleware in favour of a simpler fastapi dependency. 

(FWIW, I wrote an async only fastapi sqla lib which illustrates my thought-> https://github.com/hadrien/fastapi-async-sqla )

arththebird:
> FWIW, when I started this lib, I initially wrote a fastapi dependency to handle the sqla session.
> 
> But there was a bug/flaw in fastapi ([fastapi/fastapi#869](https://github.com/fastapi/fastapi/issues/869)).
> 
> Therefore, @vicrep and I went with a middleware instead.
> 
> Today, that fastapi bug/flaw has been fixed: I strongly recommend ditching the middleware in favour of a simpler fastapi dependency.
> 
> (FWIW, I wrote an async only fastapi sqla lib which illustrates my thought-> https://github.com/hadrien/fastapi-async-sqla )

Thanks for flagging, we'll look into this!

PR #234, opened 2024-12-03
fix: allow using python < 3.14 🥧

hadrien:
## Problem

* Current constraint on python version disallows using any python version greater or equal than 3.13

## Solution

* Set python version constraint to `>=3.9,<3.14`
* Update ci to test against python 3.10 -> 3.13

## Related

* [DIA-76532]

## Validation

* Tests pass.

[DIA-76532]: https://dialoguemd.atlassian.net/browse/DIA-76532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -33,24 +33,24 @@ workflows:
...
           context: code-validator
-      
+
       - test:
           name: 🧪 test py<<matrix.python_version>> sqlalchemy-<<matrix.sqlalchemy_version >> pydantic-<<matrix.pydantic_version>> <<matrix.asyncpg>> <<matrix.aws_rds_iam>>
           context: code-validator
           matrix:
             parameters:
-              python_version: ["3.9", "3.10", "3.11", "3.12"]
+              python_version: ["3.10", "3.11", "3.12", "3.13"]
```
arththebird, line 48: Should we keep testing with python 3.9 if we still support it?

hadrien: The test matrix would be too big afaik

hadrien: So should we keep supporting 3.9? I would say no.

arththebird: > The test matrix would be too big afaik

Too big for CCI to load you mean?

> So should we keep supporting 3.9? I would say no.

Python 3.9 EOL is on 31 Oct 2025, so there's technically almost a year of support for it. I guess it's up to us to decide if we want support still valid python versions or not. On Dialogue side, looks like we only one or two repo that still use 3.9, but we support it in all our libs.

hadrien: I put it back and cci is actually :ok: with it

hadrien:
I unfortunately cannot merge this because [codecov has failed processing the last commit](https://app.codecov.io/gh/dialoguemd/fastapi-sqla/commit/706ad54c8e4cf6d9483812193d5aa32932fdb973)

image

As it passed [yesterday](https://github.com/dialoguemd/fastapi-sqla/pull/234#issuecomment-2515375669) I'll check if it is related to the number of steps in the ci job by removing 3.9 from the matrix FYI @arththebird.

hadrien:
well, it seems it was just yet another spurious error from codecov ... 🚀

PR #266, opened 2025-02-06
fix(fixtures): add clean up post-conditions to pytest fixture async_sqla_connection

conste11ations:
## Related

* [DIA-XXXXX]

### DIFF
--- a/fastapi_sqla/_pytest_plugin.py
+++ b/fastapi_sqla/_pytest_plugin.py
@@ -159,6 +159,8 @@ if asyncio_support:
     async def async_sqla_connection(async_engine):
         async with async_engine.connect() as connection:
             yield connection
+            await connection.close()
+            await async_engine.dispose()
 
     @fixture
     async def patch_new_engine(request, async_sqla_connection):


fastapi_sqla/_pytest_plugin.py
```diff
@@ -153,7 +153,9 @@ def async_sqlalchemy_url(db_url):
 
     @fixture
     def async_engine(async_sqlalchemy_url):
-        return create_async_engine(async_sqlalchemy_url)
+        engine = create_async_engine(async_sqlalchemy_url)
+        yield engine
+        engine.dispose()
```
thibremy-dialogue, line 158: On your notion page your context is an`async` fixture`.

The engine.dispose() will just block on this `sync fixture` context or behind the scene a future/promise is run and will be completed at some point in time ?

arththebird, line 158: Should this be awaited? IIUC it's an async function

thibremy-dialogue (approved):
Thank you !

arththebird (changes requested):
> I'm open to ideas on how to add a test case that can validate fixture teardown code that usually happens outside of a test case - otherwise there is the Notion doc's findings. I can also demonstrate that this change + removing the short-term workaround on the other PR = solves the connection leak.

Can we add a test that replicates what was happening when you encountered the issue and confirm that this indeed fixes the issue?

conste11ations:
Hi @thibremy-dialogue @arththebird, I'm parking this temporarily for now:

As it turns out, `document-center` uses a hardcoded version of `fastapi-sqla` at `2.10.1`. I've got to get it up to `3.4.8` (document-center tests fail on this version right now) before I can demonstrate that the change I am proposing (and what sqlalchemy recommends as best practice) will resolve the original problem.  

@arththebird just putting down notes here - we had a chat and we agreed that, in place of a test, I will link this PR branch as a dependency to `document-center` and demonstrate that the issue is resolved with this PR. I'll provide video evidence of that once I get `document-center` to use the latest version of `fastapi-sqla`!

fastapi_sqla/_pytest_plugin.py
hadrien: Change the fixture to an async one and await the dispose call:

```python
    @fixture
    async def async_engine(async_sqlalchemy_url):
        engine = create_async_engine(async_sqlalchemy_url)
        yield engine
        await engine.dispose()
```

hadrien: indeed, it is. https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncEngine.dispose

PR #299, opened 2025-04-15
feat(pytest): support startup in tests

arththebird:
## Problem

In the pytest plugin, we patch the `sqla.new_engine` function (and its async equivalent) to return a `Connection` object. This causes issue when running the `startup` function to initialize the session factories, since it expects an engine and uses it to validate that it can connect to the database.

## Solution

Update the `sqla` and `asyncsqla` module to properly represent that `new_engine` can return either an Engine or a Connection. Update their `startup` function to properly handle both classes.

### Additional changes
- Properly type the pytest plugin to ease working on it
- Align `patch_new_engine` and `patch_new_async_engine` fixture implementation to close the transaction on tear down.
- Patch the `new_engine` and `new_async_engine` methods directly in the pytest plugin instead of patching the sqlalchemy function, so it's easier to reason about.
- Update the fixture request tree so that the `session_factory` fixture always returns session bound to the main test connection.
- Update the `async_session_factory` fixture to set `expire_on_commit=False` in order to allow accessing attributes after a commit.

## Validation

Tests are added to confirm that the startup function can safely be called and that it works with `open_session`. ✅ 

Tested with [document-center](https://github.com/dialoguemd/document-center/pull/904) to unblock the upgrade to 3.x ✅

### DIFF
--- a/tests/test_open_session.py
+++ b/tests/test_open_session.py
@@ -28,19 +28,17 @@ def test_table_cls():
 
 
 @fixture
-def startup(sqla_connection):
-    from fastapi_sqla.sqla import _DEFAULT_SESSION_KEY, _session_factories, startup
+def startup(session):
+    from fastapi_sqla.sqla import startup
 
     startup()
-    _session_factories[_DEFAULT_SESSION_KEY].configure(bind=sqla_connection)
 
 
 @fixture
-async def async_startup(async_session_key, async_sqla_connection):
-    from fastapi_sqla.async_sqla import _async_session_factories, startup
+async def async_startup(async_session, async_session_key):
+    from fastapi_sqla.async_sqla import startup
 
     await startup(async_session_key)
-    _async_session_factories[async_session_key].configure(bind=async_sqla_connection)
 
 
 @mark.sqlalchemy("1.4")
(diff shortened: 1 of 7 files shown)


tests/test_open_session.py
```diff
@@ -28,19 +28,17 @@ class TestTable(Base):
 
 
 @fixture
-def startup(sqla_connection):
-    from fastapi_sqla.sqla import _DEFAULT_SESSION_KEY, _session_factories, startup
+def startup(session):
+    from fastapi_sqla.sqla import startup
 
     startup()
-    _session_factories[_DEFAULT_SESSION_KEY].configure(bind=sqla_connection)
```
arththebird, line 35: This is no longer necessary now that the issue is fixed

tests/test_pytest_plugin.py
```diff
@@ -89,36 +89,36 @@ async def test_async_session_fixture_dont_patch_engine_writes_in_db(
 
 @mark.sqlalchemy("1.4")
 def test_all_opened_sessions_are_within_the_same_transaction(
-    sqla_connection, session, session_factory, singer_cls
+    session, session_factory, singer_cls
 ):
     session.add(singer_cls(id=1, name="Bob Marley", country="Jamaica"))
     session.commit()
 
-    other_session = session_factory(bind=sqla_connection)
```
arththebird, line 97: The session factory now automatically binds the sqla_connection, so this is no longer necesary

marcoriopel (approved):
lgtm fwiw

PR #398, opened 2026-03-24
feat: set hide params to true postgres engine [DIA-86641]

erikkostashuk:
## Problem
- when postgres throws an error, some services will log the params from the query which [gets logged in datadog](https://app.datadoghq.com/logs?query=service%3A%28engage-app%20OR%20engage-engage-to-pinpoint-worker%20OR%20engage-pull-pinpoint-events-worker%20OR%20engage-user-events-worker%20OR%20engage-job-scheduled-messages-dispatcher%20OR%20engage-load-source-identities-worker%29%20status%3Aerror%20IntegrityError&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&event=AwAAAZzN3q1ODKAU9QAAABhBWnpOM3JJTEFBQmhMckhVWkpxTkN3QXUAAAAkMTE5Y2NlMTUtZDQxOC00OGU4LWI1NGUtMjQzYzcyODAwZmFiAAnzVg&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=time%2Casc&viz=stream&from_ts=1772976953588&to_ts=1774272953588&live=true)
- [another example of Scribe logging PHI/PII](https://app.datadoghq.com/logs?query=IntegrityError&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&event=AwAAAZ0mFIIuh1J2rgAAABhBWjBtRklxYUFBQ29ISWg5VHA5bjZ3QVUAAAAkZjE5ZDI2MTQtOGE2OC00NjMwLTkyMWEtMmEyYjJkNGFiNjZiAAAaug&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=time%2Cdesc&viz=stream&from_ts=1772976953588&to_ts=1774272953588&live=true)

## Solution
- to prevent PHI/PII from getting logged in datadog, we should enable `hide_parameters` across all services that are using postgres
- we have enabled this in engage already, which results in the sql params being redacted when there is an error: [see example log with this param as true](https://app.datadoghq.com/logs?query=service%3A%28engage-app%20OR%20engage-engage-to-pinpoint-worker%20OR%20engage-pull-pinpoint-events-worker%20OR%20engage-user-events-worker%20OR%20engage-job-scheduled-messages-dispatcher%20OR%20engage-load-source-identities-worker%29%20status%3Aerror%20IntegrityError&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&event=AwAAAZ0arYT5VZGajgAAABhBWjBhclpJU0FBQ21pUDVuS0JycEJnRDQAAAAkMDE5ZDFhYmMtNmVkMC00NjNkLWFhOTctZjk5ODg1NDkwMTNmAApCzw&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=time%2Cdesc&viz=stream&from_ts=1772976953588&to_ts=1774272953588&live=true)

Changes:                                                                                                                                                                
  - Added _EngineConfig pydantic model with hide_parameters: bool = True default and v1/v2 pydantic branching
  - Extracted shared _get_engine_config() used by both sync and async engines                                                                                                  
  - Added parametrized tests for sync and async engines                               
  - Documented in README how to opt out

## Related
- SQL docs on enabling this with example: https://docs.sqlalchemy.org/en/21/core/engines.html#hiding-parameters
- https://github.com/dialoguemd/charts/pull/757#issuecomment-4112691741
- good related thread: https://dialoguemd.slack.com/archives/C06P5H4NRBM/p1774642813877189

* [DIA-86641]

## Validation

Testing in timekeeper
Screenshot 2026-03-24 at 1 32 53 PM

```
poetry add git+https://github.com/dialoguemd/fastapi-sqla.git#feat/hide-sql-params-DIA-86641
```

```
poetry run python -c "import os; os.environ['SQLALCHEMY_URL'] = 'sqlite:///test.db'; from fastapi_sqla.sqla import new_engine; e = new_engine(); print('hide_parameters:', e.engine.hide_parameters)"
```

- https://github.com/dialoguemd/timekeeper/pull/1697

[DIA-86641]: https://dialoguemd.atlassian.net/browse/DIA-86641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fastapi_sqla/_pytest_plugin.py
+++ b/fastapi_sqla/_pytest_plugin.py
@@ -58,7 +58,7 @@ def db_url(db_host: str, db_user: str) -> str:
 
 @fixture(scope="session")
 def engine(db_url: str) -> Engine:
-    return create_engine(db_url, hide_parameters=True)
+    return create_engine(db_url)
 
 
 @fixture(scope="session")
@@ -159,7 +159,7 @@ if asyncio_support:
 
     @fixture
     def async_engine(async_sqlalchemy_url: str) -> AsyncEngine:
-        return create_async_engine(async_sqlalchemy_url, hide_parameters=True)
+        return create_async_engine(async_sqlalchemy_url)
 
     @fixture
     async def async_sqla_connection(
--- a/fastapi_sqla/_pytest_plugin.py
+++ b/fastapi_sqla/_pytest_plugin.py
@@ -58,7 +58,7 @@ def db_url(db_host: str, db_user: str) -> str:
 
 @fixture(scope="session")
 def engine(db_url: str) -> Engine:
-    return create_engine(db_url)
+    return create_engine(db_url, hide_parameters=True)
 
 
 @fixture(scope="session")
@@ -159,7 +159,7 @@ if asyncio_support:
 
     @fixture
     def async_engine(async_sqlalchemy_url: str) -> AsyncEngine:
-        return create_async_engine(async_sqlalchemy_url)
+        return create_async_engine(async_sqlalchemy_url, hide_parameters=True)
 
     @fixture
     async def async_sqla_connection(
--- a/fastapi_sqla/sqla.py
+++ b/fastapi_sqla/sqla.py
@@ -75,8 +75,7 @@ def _get_engine_config(
         if k.startswith(envvar_prefix)
     }
     config = _EngineConfig(**overrides)  # type: ignore[arg-type]
-    coerced = config.dict()
-    for param, value in coerced.items():
+    for param, value in config.dict().items():
         lowercase_env[f"{envvar_prefix}{param}"] = value
 
     return lowercase_env
--- a/fastapi_sqla/sqla.py
+++ b/fastapi_sqla/sqla.py
@@ -9,7 +9,6 @@ from fastapi import Depends, Request, Response
 from fastapi.concurrency import contextmanager_in_threadpool
 from fastapi.responses import PlainTextResponse
 from pydantic import BaseModel
-from pydantic import __version__ as pydantic_version
 from sqlalchemy import engine_from_config, text
 from sqlalchemy.engine import Connection, Engine
 from sqlalchemy.ext.declarative import DeferredReflection
@@ -40,26 +39,13 @@ _REQUEST_SESSION_KEY = "fastapi_sqla_session"
 _session_factories: dict[str, sessionmaker] = {}
 
 
-_pydantic_major = int(pydantic_version.split(".")[0])
+class _EngineConfig(BaseModel):
+    """Engine configuration with typed defaults and bool coercion."""
 
-if _pydantic_major == 2:
-    from pydantic import ConfigDict
+    hide_parameters: bool = True
 
-    class _EngineConfig(BaseModel):
-        """Engine configuration with typed defaults and bool coercion."""
-
-        model_config = ConfigDict(extra="allow")
-        hide_parameters: bool = True
-
-else:
-
-    class _EngineConfig(BaseModel):  # type: ignore[no-redef]
-        """Engine configuration with typed defaults and bool coercion."""
-
-        hide_parameters: bool = True
-
-        class Config:
-            extra = "allow"
+    class Config:
+        extra = "allow"
 
 
 class Base(DeclarativeBase, DeferredReflection):
@@ -89,7 +75,7 @@ def _get_engine_config(
         if k.startswith(envvar_prefix)
     }
     config = _EngineConfig(**overrides)  # type: ignore[arg-type]
-    coerced = config.model_dump() if _pydantic_major == 2 else config.dict()
+    coerced = config.dict()
     for param, value in coerced.items():
         lowercase_env[f"{envvar_prefix}{param}"] = value
 
--- a/fastapi_sqla/sqla.py
+++ b/fastapi_sqla/sqla.py
@@ -104,7 +104,7 @@ def _get_engine_config(
     lowercase_env.pop(f"{envvar_prefix}warn_20", None)
 
     overrides = {
-        k[len(envvar_prefix):]: v
+        k[len(envvar_prefix) :]: v
         for k, v in lowercase_env.items()
         if k.startswith(envvar_prefix)
     }
--- a/fastapi_sqla/sqla.py
+++ b/fastapi_sqla/sqla.py
@@ -9,7 +9,6 @@ from fastapi import Depends, Request, Response
 from fastapi.concurrency import contextmanager_in_threadpool
 from fastapi.responses import PlainTextResponse
 from pydantic import BaseModel
-from pydantic import __version__ as pydantic_version
 from sqlalchemy import engine_from_config, text
 from sqlalchemy.engine import Connection, Engine
 from sqlalchemy.ext.declarative import DeferredReflection
@@ -19,14 +18,17 @@ from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 from fastapi_sqla import aws_aurora_support, aws_rds_iam_support
 
-_pydantic_major = int(pydantic_version.split(".")[0])
-
 try:
     from sqlalchemy.orm import DeclarativeBase
 except ImportError:
     from sqlalchemy.ext.declarative import declarative_base
 
     DeclarativeBase = declarative_base()  # type: ignore
+
+try:
+    from pydantic import ConfigDict as _ConfigDict
+except ImportError:
+    _ConfigDict = None  # type: ignore[assignment,misc]
 
 try:
     from sqlmodel import Session as SqlaSession  # type: ignore
@@ -42,33 +44,15 @@ _REQUEST_SESSION_KEY = "fastapi_sqla_session"
 _session_factories: dict[str, sessionmaker] = {}
 
 
-def _coerce_bool_strings(data: dict) -> dict:
-    """Coerce 'true'/'false' strings to bool in a dict."""
-    coerced = {}
-    for k, v in data.items():
-        if isinstance(v, str) and v.lower() in ("true", "false"):
-            coerced[k] = v.lower() == "true"
-        else:
-            coerced[k] = v
-    return coerced
-
-
-if _pydantic_major == 2:
-    from pydantic import model_validator
+if _ConfigDict is not None:
 
     class _EngineConfig(BaseModel):
         """Engine configuration with typed defaults and bool coercion."""
 
-        model_config = {"extra": "allow"}
+        model_config = _ConfigDict(extra="allow")
         hide_parameters: bool = True
 
-        @model_validator(mode="before")
-        @classmethod
-        def coerce_booleans(cls, data):
-            return _coerce_bool_strings(data)
-
 else:
-    from pydantic import root_validator
 
     class _EngineConfig(BaseModel):  # type: ignore[no-redef]
         """Engine configuration with typed defaults and bool coercion."""
@@ -77,10 +61,6 @@ else:
 
         class Config:
             extra = "allow"
-
-        @root_validator(pre=True, allow_reuse=True)
-        def coerce_booleans(cls, values):  # noqa: N805
-            return _coerce_bool_strings(values)
 
 
 class Base(DeclarativeBase, DeferredReflection):
@@ -110,7 +90,7 @@ def _get_engine_config(
         if k.startswith(envvar_prefix)
     }
     config = _EngineConfig(**overrides)  # type: ignore[arg-type]
-    coerced = config.model_dump() if _pydantic_major == 2 else config.dict()
+    coerced = config.model_dump() if hasattr(config, "model_dump") else config.dict()
     for param, value in coerced.items():
         lowercase_env[f"{envvar_prefix}{param}"] = value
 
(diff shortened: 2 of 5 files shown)


freyjajok (approved):
LGTM! Just a small nitpick

fastapi_sqla/_pytest_plugin.py
```diff
@@ -58,7 +58,7 @@ def db_url(db_host: str, db_user: str) -> str:
 
 @fixture(scope="session")
 def engine(db_url: str) -> Engine:
-    return create_engine(db_url)
+    return create_engine(db_url, hide_parameters=True)
```
arththebird: Why do we want to hide params in tests?

fastapi_sqla/sqla.py
```diff
@@ -50,11 +50,36 @@ def get_envvar_prefix(key: str) -> str:
...
+    lowercase_env.pop(f"{envvar_prefix}warn_20", None)
+    for param, default in defaults.items():
+        env_key = f"{envvar_prefix}{param}"
+        if env_key not in lowercase_env:
+            lowercase_env[env_key] = default
+        elif isinstance(default, bool):
+            env_val = lowercase_env[env_key]
+            lowercase_env[env_key] = (
+                isinstance(env_val, str) and env_val.lower() == "true"
+            )
```
arththebird: Why do we convert the env var values to bool here? The env var doesn't work OOTB?

arththebird: And if it doesn't work OOTB, should we instead parse all boolean strings as bool so all params expecting to be bool do work, and not just the defaults?

arththebird: Reading [here](https://docs.sqlalchemy.org/en/21/core/engines.html#sqlalchemy.engine_from_config:~:text=A%20select%20set%20of%20keyword%20arguments%20will%20be%20%E2%80%9Ccoerced%E2%80%9D%20to%20their%20expected%20type%20based%20on%20string%20values.%20The%20set%20of%20arguments%20is%20extensible%20per%2Ddialect%20using%20the%20engine_config_types%20accessor.), this should work OOTB:

> A select set of keyword arguments will be “coerced” to their expected type based on string values. The set of arguments is extensible per-dialect using the engine_config_types accessor.

fastapi_sqla/_pytest_plugin.py
erikkostashuk: Good call, reverted that

fastapi_sqla/sqla.py
erikkostashuk: I don't think `hide_parameters` will get automatically coerced to the correct type? The list that gets coerced is [here](https://github.com/sqlalchemy/sqlalchemy/blob/f244d5c3a92bffb93bfbaa279734a3a508da7bc3/lib/sqlalchemy/engine/default.py#L186)

tests/test_sqla_startup.py
```diff
@@ -190,3 +190,65 @@ def now():
...
+    assert config["sqlalchemy_echo"] == "debug"
+
+
+def test_get_engine_config_non_bool_default_not_set(monkeypatch):
+    from fastapi_sqla.sqla import _get_engine_config
+
+    monkeypatch.delenv("sqlalchemy_echo", raising=False)
+    config = _get_engine_config("sqlalchemy_", defaults={"echo": "info"})
+
+    assert config["sqlalchemy_echo"] == "info"
```
sachasmart: Can we parameterize these two tests?

tests/test_sqla_startup.py
```diff
@@ -190,3 +190,65 @@ def now():
...
+
+
+def test_new_engine_hide_parameters_can_be_disabled(monkeypatch):
+    from fastapi_sqla.sqla import new_engine
+
+    monkeypatch.setenv("sqlalchemy_hide_parameters", "false")
+
+    engine_or_conn = new_engine()
+
+    assert engine_or_conn.engine.hide_parameters is False
```
sachasmart: Lets parameterize this as well

fastapi_sqla/sqla.py
```diff
@@ -50,11 +50,36 @@ def get_envvar_prefix(key: str) -> str:
     return envvar_prefix
 
 
+_ENGINE_DEFAULTS = {
+    "hide_parameters": True,
+}
+
```
sachasmart: Could we use a pydantic model here? I think this could help with some of the type coercion as well

Possibly something like:

```python
class EngineConfig(BaseModel):
    hide_parameters: bool = True
    ....

    @field_validator("*", mode="before")
    @classmethod
    def coerce_booleans(cls, value):
        .... your logic
```
Then we wouldn't need this type check (e.g., instance(..))

Let me know if I'm missing something!

sachasmart: Also, super nit but can you move this up to line 39

sachasmart (commented):
Nice work, Erik! Let me know if I'm (likely) missing something

tests/test_sqla_startup.py
erikkostashuk: done

tests/test_sqla_startup.py
erikkostashuk: done

tests/test_sqla_startup.py
```diff
@@ -190,3 +190,68 @@ def now():
...
+        monkeypatch.delenv("sqlalchemy_hide_parameters", raising=False)
+
+    engine_or_conn = new_engine()
+
+    assert engine_or_conn.engine.hide_parameters is expected
+
+
+@mark.require_asyncpg
+@mark.sqlalchemy("1.4")
+async def test_new_async_engine_hides_parameters_by_default(async_session_key):
```
arththebird: Should we parametrize these tests as well?

fastapi_sqla/sqla.py
arththebird: Ah you're right, that's sad 😞 In this case, I do like [Sacha's suggestion](https://github.com/dialoguemd/fastapi-sqla/pull/398#discussion_r3012085493) for this so we parse all boolean strings as bool and not just the ones in `_ENGINE_DEFAULTS`

tests/test_sqla_startup.py
erikkostashuk: yes we should, done

fastapi_sqla/sqla.py
erikkostashuk: yea super weird case, I wonder why they only coerce some args

I like it too, implementing that now

fastapi_sqla/sqla.py
erikkostashuk: done, the pydantic model should actually handle the coercion without needing a field validator

fastapi_sqla/sqla.py
```diff
@@ -38,6 +39,15 @@
...
 
 
+class _EngineConfig(BaseModel):
+    """Engine configuration with typed defaults and bool coercion."""
+
+    hide_parameters: bool = True
+
+    class Config:
+        extra = "allow"
+
```
sachasmart, line 49: Please don't hate me, but given this project is on pydantic v2, can we update this to:

```python
.. import from pydantic.whatever
model_config = ConfigDict(extra="allow")
```

fastapi_sqla/sqla.py
```diff
@@ -50,11 +60,32 @@ def get_envvar_prefix(key: str) -> str:
...
+    }
+    lowercase_env.pop(f"{envvar_prefix}warn_20", None)
+
+    overrides = {
+        k[len(envvar_prefix) :]: v
+        for k, v in lowercase_env.items()
+        if k.startswith(envvar_prefix)
+    }
+    config = _EngineConfig(**overrides)  # type: ignore[arg-type]
+    coerced = config.dict()
```
sachasmart: Lame but can we dump to a json here? `.model_dump(mode=json)` or `.model_dump_json`

fastapi_sqla/sqla.py
```diff
@@ -50,11 +60,32 @@ def get_envvar_prefix(key: str) -> str:
     return envvar_prefix
 
 
+def _get_engine_config(
+    envvar_prefix: str,
```
sachasmart, line 64: Sorry why can't we use `BaseSettings` with:

```python
class Config(BaseSettings):
    model_config = SettingsConfigDict(env_prefix="fastapi_sqla")
```

fastapi_sqla/sqla.py
```diff
@@ -38,6 +39,15 @@
 _session_factories: dict[str, sessionmaker] = {}
 
 
+class _EngineConfig(BaseModel):
```
sachasmart, line 42: sorry can't we use `BaseSettings` here with a prefix? Then this will read anything with a prefix of `fastapi_sqla_`

sachasmart (commented):
Sorry just comments about reading from the environment

fastapi_sqla/sqla.py
erikkostashuk: It got removed here: https://github.com/dialoguemd/fastapi-sqla/blob/feat/hide-sql-params-DIA-86641/CHANGELOG.md#v290-2023-07-31

fastapi_sqla/sqla.py
erikkostashuk: we need to support v1 and v2 pydantic in this repo, using branching now to use the specific methods for each verison

fastapi_sqla/sqla.py
erikkostashuk: same here, addressed with branching for supporting both pydantic versions

fastapi_sqla/sqla.py
erikkostashuk: same as: https://github.com/dialoguemd/fastapi-sqla/pull/398#discussion_r3030281685

fastapi_sqla/sqla.py
```diff
@@ -38,6 +40,28 @@
 _session_factories: dict[str, sessionmaker] = {}
 
 
+_pydantic_major = int(pydantic_version.split(".")[0])
+
+if _pydantic_major == 2:
```
marcoriopel: I think the Config class is only deprecated in v2 but still supported. You don't need to do this if else block

fastapi_sqla/sqla.py
```diff
@@ -50,11 +74,32 @@ def get_envvar_prefix(key: str) -> str:
...
+    }
+    lowercase_env.pop(f"{envvar_prefix}warn_20", None)
+
+    overrides = {
+        k[len(envvar_prefix) :]: v
+        for k, v in lowercase_env.items()
+        if k.startswith(envvar_prefix)
+    }
+    config = _EngineConfig(**overrides)  # type: ignore[arg-type]
+    coerced = config.model_dump() if _pydantic_major == 2 else config.dict()
```
marcoriopel: Same than [here](https://github.com/dialoguemd/fastapi-sqla/pull/398/changes#r3053251478). I would simplify the code and when we get to not supporting v1 anymore we change it for the v2 way.

tests/test_sqla_startup.py
```diff
@@ -190,3 +190,52 @@ def now():
...
+
+
+@mark.parametrize(
+    "env_value, expected",
+    [
+        (None, True),
+        ("false", False),
+    ],
+)
+def test_new_engine_hide_parameters(monkeypatch, env_value, expected):
```
marcoriopel: Having a parametrize with an if else block here is actually more code than having the two tests separated.

tests/test_sqla_startup.py
```diff
@@ -190,3 +190,52 @@ def now():
...
+@mark.require_asyncpg
+@mark.sqlalchemy("1.4")
+@mark.parametrize(
+    "env_value, expected",
+    [
+        (None, True),
+        ("false", False),
+    ],
+)
+async def test_new_async_engine_hide_parameters(
```
marcoriopel: Same than [this](https://github.com/dialoguemd/fastapi-sqla/pull/398/changes#r3053268283)

fastapi_sqla/sqla.py
erikkostashuk: updated

fastapi_sqla/sqla.py
erikkostashuk: updated

tests/test_sqla_startup.py
erikkostashuk: updated

tests/test_sqla_startup.py
erikkostashuk: updated

fastapi_sqla/sqla.py
```diff
@@ -50,11 +60,32 @@ def get_envvar_prefix(key: str) -> str:
...
+    lowercase_env.pop(f"{envvar_prefix}warn_20", None)
+
+    overrides = {
+        k[len(envvar_prefix) :]: v
+        for k, v in lowercase_env.items()
+        if k.startswith(envvar_prefix)
+    }
+    config = _EngineConfig(**overrides)  # type: ignore[arg-type]
+    coerced = config.dict()
+    for param, value in coerced.items():
```
marcoriopel:
```suggestion
    for param, value in config.dict().items():
```

marcoriopel (approved):

PR #407, opened 2026-05-11
feat(deps): remove support for python 3.9

arththebird:
## Problem

Python 3.9 has been EOL since October 2025. We still support it and this starting to cause issue updating our deps since they dropped support for 3.9.

All our core repos are not on 3.9 anymore.

## Solution

Drop support for python 3.9 and fix all lint error following upgrade of minimum supported python version.

## Validation

All test still pass ✅


Makefile
```diff
@@ -0,0 +1,9 @@
+test:
```
arththebird, line 1: Unrelated, but was as PITA when working in this repo

pyproject.toml
```diff
@@ -162,9 +161,6 @@ addopts = """
 # https://nedbatchelder.com/blog/201810/why_warnings_is_mysterious.html
 filterwarnings = ["error:.*removed in version 2.0.*:"]
 
-[tool.pytest-watch.run]
```
arththebird, line 165: Was removed a while ago

marcoriopel:
Should this be considered a breaking change here as it's a public repo so other users could still rely on 3.9?

arththebird:
> Should this be considered a breaking change here as it's a public repo so other users could still rely on 3.9?

This is a good question. I personally followed what most of our upstream deps have done, which is not consider it a breaking change. 🤷‍♂️

marcoriopel (approved):

PR #410, opened 2026-05-13
feat: add Pydantic v1 compatibility layer (WIP)

thanethomson:
## Problem

At present, Fastapi-SQLA tries to [autodetect the currently installed version of Pydantic](https://github.com/dialoguemd/fastapi-sqla/blob/982ba42703027cebd2da9088f25f0ca5f68aa051/fastapi_sqla/models.py#L8), using the relevant base model based on the currently installed version.

If, however, a dependent project wants to leverage Pydantic's v1 compatibility layer (e.g. they're undergoing a large, complex migration), this autodetection could cause breaking changes to other projects that depend on the dependent project.

## Solution

This is one possible solution of several, which involves duplicating all relevant functionality in the core library into a `fastapi_sqla.v1` compatibility layer that:
1. Uses Pydantic v1 models if v1 is installed.
2. Uses Pydantic v2's v1 compatibility layer if v2 is installed.

The idea here would be to support this until such time that Fastapi-SQLA upgrades to Pydantic v3, where it's currently anticipated that Pydantic will drop support for its v1 compatibility layer.

Commit should be independently reviewable. See also [this section in the README](https://github.com/dialoguemd/fastapi-sqla/blob/thane/pydanticv1-compat/README.md#pydantic-v1-compatibility) (added in this PR) regarding the compatibility layer and its usage.

#### Option 1: Use an old version of Fastapi-SQLA in dependent projects

This is... okay-ish. But obviously dependent projects lose the benefits of more recent features and dependency updates.

#### Option 2: Environment variable

One could let a `FASTAPI_SQLA_PYDANTIC_VERSION` environment variable dictate which version of Pydantic to use. This would require a much smaller code change than this PR, but would also be fairly brittle from a usage perspective (e.g. what if one forgets to set the environment variable in one's deployment environment?).

#### Option 3: `pyproject.toml` configuration

This would also potentially be brittle, because it would require that Fastapi-SQLA traverse the project hierarchy to leverage the main project's `pyproject.toml` and read a `fastapi-sqla` configuration block to get the version of Pydantic it needs to use.

## Related

* [PS-1327](https://dialoguemd.atlassian.net/browse/PS-1327)

## Validation

I'm currently in the process of testing out using this new compatibility layer on our internal projects/repos and will report back here (and mark as ready for review) when done.

Results of `make diff:v1` introduced in this PR:

```diff
diff --git a/fastapi_sqla/models.py b/fastapi_sqla/v1/models.py
index a91389f..dd60756 100644
--- a/fastapi_sqla/models.py
+++ b/fastapi_sqla/v1/models.py
@@ -1,13 +1,14 @@
 from typing import Generic, TypeVar

-from pydantic import BaseModel, Field
 from pydantic import __version__ as pydantic_version

 major, _, _ = (int(v) for v in pydantic_version.split("."))
 is_pydantic2 = major == 2
 if is_pydantic2:
-    GenericModel = BaseModel
+    from pydantic.v1 import BaseModel, Field
+    from pydantic.v1.generics import GenericModel  # type:ignore
 else:
+    from pydantic import BaseModel, Field  # type:ignore
     from pydantic.generics import GenericModel  # type:ignore


diff --git a/fastapi_sqla/pagination.py b/fastapi_sqla/v1/pagination.py
index c1f6807..4148a9f 100644
--- a/fastapi_sqla/pagination.py
+++ b/fastapi_sqla/v1/pagination.py
@@ -7,8 +7,8 @@ from fastapi import Depends, Query
 from sqlalchemy.orm import Query as LegacyQuery
 from sqlalchemy.sql import Select, func, select

-from fastapi_sqla.models import Meta, Page
 from fastapi_sqla.sqla import _DEFAULT_SESSION_KEY, SessionDependency, SqlaSession
+from fastapi_sqla.v1.models import Meta, Page

 DbQuery = LegacyQuery | Select
 QueryCountDependency = Callable[..., int]
@@ -19,13 +19,6 @@ PaginateDependency = DefaultDependency | WithQueryCountDependency


 def default_query_count(session: SqlaSession, query: DbQuery) -> int:
-    """Default function used to count items returned by a query.
-
-    It is slower than a manually written query could be: It runs the query in a
-    subquery, and count the number of elements returned.
-
-    See https://gist.github.com/hest/8798884
-    """
     if isinstance(query, LegacyQuery):
         result = query.count()

@@ -52,7 +45,6 @@ def paginate_query(
     limit: int,
     scalars: bool = True,
 ) -> Page:  # pragma: no cover
-    """Dispatch on registered functions based on `query` type"""
     raise NotImplementedError(f"no paginate_query registered for type {type(query)!r}")


diff --git a/fastapi_sqla/async_pagination.py b/fastapi_sqla/v1/async_pagination.py
index a82a4f9..70a8071 100644
--- a/fastapi_sqla/async_pagination.py
+++ b/fastapi_sqla/v1/async_pagination.py
@@ -6,8 +6,8 @@ from fastapi import Depends, Query
 from sqlalchemy.sql import Select, func, select

 from fastapi_sqla.async_sqla import AsyncSessionDependency, SqlaAsyncSession
-from fastapi_sqla.models import Meta, Page
 from fastapi_sqla.sqla import _DEFAULT_SESSION_KEY
+from fastapi_sqla.v1.models import Meta, Page

 QueryCountDependency = Callable[..., Awaitable[int]]
 AsyncPaginateSignature = Callable[[Select, bool | None], Awaitable[Page]]
```

[PS-1327]: https://dialoguemd.atlassian.net/browse/PS-1327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,7 @@
+diff\:v1:
+	@for f in models.py pagination.py async_pagination.py; do \
+		git diff --no-index fastapi_sqla/$$f fastapi_sqla/v1/$$f || true; \
+	done
 test:
 	poetry run pytest
 test\:cov:
--- a/README.md
+++ b/README.md
@@ -841,6 +841,20 @@ $ poetry run pytest
 $ poetry run tox
 ```
 
+## Pydantic v1 compatibility layer diff
+
+A `diff:v1` target is provided in the `Makefile` to help show the diff between
+the v1 compatibility layer and the files the compatibility layer aims to
+replicate.
+
+```bash
+make diff:v1
+```
+
+When making changes to the primary types/functionality, use `make diff:v1` to
+ensure that meaningful changes are propagated into the compatibility layer while
+support for Pydantic v1 is still provided.
+
 [`sqlalchemy.create_engine`]: https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine
 [`Query.count`]: https://docs.sqlalchemy.org/en/20/orm/queryguide/query.html#sqlalchemy.orm.Query.count
 [pytest options]: https://docs.pytest.org/en/stable/reference.html#confval-usefixtures
--- a/README.md
+++ b/README.md
@@ -835,7 +835,7 @@ $ poetry install --all-extras
 $ poetry run pytest
 ```
 
-#### Runing tests on multiple environments
+#### Running tests on multiple environments
 
 ```bash
 $ poetry run tox
--- a/README.md
+++ b/README.md
@@ -582,6 +582,68 @@ async def async_all_users_alt(
     return await paginate(select(User))
 ```
 
+# Pydantic v1 compatibility
+
+If you are migrating a project from Pydantic v1 to v2 and need time to update your
+response models, `fastapi_sqla.v1` provides a drop-in replacement for the pagination
+helpers that produces Pydantic v1-compatible `Page`, `Item`, and `Collection` types.
+
+Replace your `fastapi_sqla` imports with `fastapi_sqla.v1`:
+
+```python
+from fastapi import APIRouter
+from fastapi_sqla import Base
+from fastapi_sqla.v1 import Page, Paginate, Session, setup
+from sqlalchemy import select
+
+# Use pydantic v1 BaseModel (or pydantic.v1 when running pydantic v2)
+try:
+    from pydantic.v1 import BaseModel  # pydantic v2
+except ImportError:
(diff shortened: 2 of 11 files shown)


thanethomson:
This PR was part of a PoC towards upgrading ER to Pydantic v2, but ultimately due to ER's dependence on FastAPI, this approach won't work out.

